PR #378, opened 2015-01-25
chromium: version bumped to 40.0.2214.91

Ratler:
- Use bundled version of protobuf or syncing will cause chromium to seg fault
- Added Lunar Linux API keys for the various Google API's chromium can utilize
- Removed unnecessary third part packages cleanup in PRE_BUILD
- Compile and install chromedriver

### DIFF
--- a/web/chromium/BUILD
+++ b/web/chromium/BUILD
@@ -1,4 +1,3 @@
-
 # Lets do some ffmpeg stuff
   pushd third_party/ffmpeg > /dev/null &&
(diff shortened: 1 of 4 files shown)


Ratler:
retest this please

stumbles:
Ratler.

Before you do anything with this, working on another version bump. Have been using the changes in that commit and haven't found any issues. Give me a day or two of using those changes with version 41.0.2272.3.

stumbles:
Never mind. Tried several of their 41 versions and there are issues with vp8, segfaults. Neither system or source provided fixes the problem.

So run with what we got, it works just fine from a couple of days of use.

PR #389, opened 2015-02-03
Boost: a couple of patches adding some missing includes.

stumbles:


v4hn:
Do you have any specific reason to add these patches?

If you can't name an (officially non-broken) package
that depends on these changes I'd like to leave them aside.
A lot of code I use uses diverse boost modules and I didn't
encounter any issues up to now...

stumbles:
https://svn.boost.org/trac/boost/ticket/10749

https://github.com/boostorg/signals2/pull/8

v4hn:
Ok, so I do agree that we should add the two patches you referenced.

However, I disagree with the third commit you added.
It's not boost's mistake that modules fail to "correctly" use qt4.

The correct fix for the build failures you point out are to add the
respective Q_MOC_RUN safeguards around the boost includes in the
source files of the dependent projects which are passed to qt's moc.

v4hn:
Oh, also this would fix only a small portion of all the issues with qt4 and boost.

Just run `grep -r BOOST_JOIN -l /usr/include/boost`.
This lists 32 files and qt4's moc breaks with each one of them included.

PR #395, opened 2015-02-11
libmng: bump to 2.0.2 along with a few changes to the BUILD. The second ...

stumbles:
...sedit is

to account for a make failure if jpeg is installed. I use libjpeg-turbo so that
part is not tested. The sedits are taken from;

http://sources.gentoo.org/cgi-bin/viewvc.cgi/gentoo-x86/media-libs/libmng/files/

### DIFF
--- a/graphics/libmng/BUILD
+++ b/graphics/libmng/BUILD
@@ -1,13 +1,8 @@
-(
 
-  cp unmaintained/autogen.sh . &&
-  chmod 755 autogen.sh &&
+  sedit "207i #include <jpeglib.h>" libmng_types.h &&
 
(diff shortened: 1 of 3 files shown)


stumbles:
Try again.

Florin65:
it compiles just fine for me.

stumbles:
These are the modules I know if installed will need a recompile;

qt4, qt5, devil, xine-lib and MPlayer

stumbles:
florin, you have jpeg or libjpeg-turbo? Not that it should matter because both contain jpeglib.h. Unless jpeg doesn't install its, but that seems unlikely.

PR #433, opened 2015-04-27
llvm: version bumped to 3.6.0.

Florin65:
It's needed for the new mesa-lib version (10.5.4) when it will be committed in the moonbase.
It breaks the existing mesa-lib.


stumbles:
Well it will break the current moonbase version of mesa-lib.

stumbles:
blender if fine with that version. 
clamav doesn't like any version => 3.60 including 0.98.7 (moonbase 0.98.6)

Did not try ghc.

PR #464, opened 2015-06-19
Jen

Florin65:


v4hn:
hey @Florin65, we already *have* a lz4 module around in libs/, so why do you want to add a new one?
It's a bit outdated though, so I suppose you could push an update to your current version.

PR #494, opened 2015-09-01
gforth: version bump to 0.7.3.

dagbrown:
Module untouched since 2008, I bet everyone else forgot that this even existed!


lunar-linux-b01:
SUCCESS!

PR #501, opened 2015-09-13
ufraw: depends lcms2 rather than lcms

rtlanceroad:
./configure tries to find lcms2, not lcms version 1 :>


lunar-linux-b01:
SUCCESS!

PR #502, opened 2015-09-13
firefox-bin: use http rather than ftp

rtlanceroad:
ftp is slow and sometimes hangs for a long time even 100%


lunar-linux-b01:
SUCCESS!

PR #521, opened 2015-11-08
cdrtools: version bumped to 3.02a01

fbrosson:
I've added an optional_depends on pulseaudio for the "-echo" feature of cdda2wav which is able to use pulseaudio since 3.02a01.
Thanks


lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

PR #526, opened 2015-12-08
augeas: new module, a configuration editing tool

dagbrown:
I've been playing with this for a couple of days, it seems like it could come in handy if you have lots of hosts to manage.

From the web site:
    Augeas is a configuration editing tool. It parses configuration
    files in their native formats and transforms them into a tree.
    Configuration changes are made by manipulating this tree and saving
    it back into native config files.


lunar-linux-b01:
SUCCESS!

PR #529, opened 2015-12-12
alsa-plugins: add missing dependency

rtlanceroad:
lastest alsa-plugins is trying to find speex rate plugin and speex preprocess plugins. In this case, speex is needed.

### DIFF
--- a/audio/alsa-plugins/DEPENDS
+++ b/audio/alsa-plugins/DEPENDS
@@ -1,4 +1,5 @@
 depends alsa-lib
+depends speex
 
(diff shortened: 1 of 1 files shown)


lunar-linux-b01:
SUCCESS!

audio/alsa-plugins/DEPENDS
```diff
@@ -1,4 +1,5 @@
 depends alsa-lib
+depends speex
```
Ratler: Speex should be an optional dependency, please take a look at "--with-speex={builtin|lib|no}" configure flag.

rtlanceroad:
@Ratler Thanks. Now i see. Can you add it to optional depends? I'd like to be a lazy guy at the time of replying :)

PR #540, opened 2016-01-14
go-bootstrap: a bootstrap go compiler module.

Florin65:
It is used only no previous go compiler is installed on the system. After the install of go module it is removed form the system.


lunar-linux-b01:
SUCCESS!

Florin65:
right now it's broken as the current go module.

PR #546, opened 2016-01-30
ebtables: new module (actually very old module)

dagbrown:
Apparently the latest version of libvirt needs this.  It's a tool
to build firewall rules at the Ethernet bridging level rather than
at the IP level.


lunar-linux-b01:
SUCCESS!

PR #557, opened 2016-02-09
lirc: add optional Python-3 dep

rtlanceroad:
The main webpage lists python3 as  Mandatory dependencies , but I think lunar doesn't treat it as mandatory.


v4hn:
Closed by 24c31eaaa2a5071ffe5771806144140ca262f880

PR #558, opened 2016-02-09
llvm: version bumped to 3.7.1.

Florin65:
Another llvm module which use cmake for build.
I haven't figured yet how can we use the R600 option with it.


lunar-linux-b01:
SUCCESS!

v4hn:
Is there any particular reason to switch to the cmake build?

doc/Packaging.rst still lists the autotools options for packaging.

v4hn:
Ok, I just talked to someone on the #llvm channel.
cmake is the officially recommended build system these days,
and autotools support will be removed after 3.8.

So, your request is the way to go.

I'll patch it up a bit and add a new request.

Florin65:
Great. Thank you.

PR #561, opened 2016-02-10
sun-jre: version bumped to 8u74.

Florin65:
reworked the module and bumped to the last version.


v4hn:
Heya, I get a 404 file not found with this request.
also you forgot to add JRE_SHA_VFY for x86_64.
To clean up old java installations which installed to /usr/java, please leave  the `rm -rf /usr/java` in the BUILD file for now

PR #563, opened 2016-02-12
llvm: version bump & cmake build

v4hn:
This is based on florin's bump

### DIFF
--- a/compilers/llvm/BUILD
+++ b/compilers/llvm/BUILD
@@ -1,7 +1,18 @@
-OPTS+=" --enable-shared --enable-targets=host-only" &&
+OPTS+=" -DBUILD_SHARED_LIBS=ON" &&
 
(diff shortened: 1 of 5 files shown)


Florin65:
I think the default for R600 option have to be "n" in the CONFIGURE file. Many users do not have a Radeon video card.

lunar-linux-b01:
SUCCESS!

Florin65:
It installs fine. Funny thing is that the new tarball is longer with 5 bytes than the previous compiled one. So, the automatic select of targets does a pretty good job. However with the new setup it might be compiled a R600 enabled install on systems without a Radeon GPU i guess.

v4hn:
I changed the R600 default to n.
I don't have a strong opinion on this one.

If it's otherwise fine, please merge the request.

lunar-linux-b01:
SUCCESS!

PR #564, opened 2016-02-18
Busybox: Add option to build as a static executable.

dagbrown:
Someone's going to have to try this out manually, I don't think jenkins can try out multiple options.


lunar-linux-b01:
SUCCESS!

PR #578, opened 2016-03-23
git: Added dependencies required when documentation requested.

dagbrown:
There doesn't seem to be an awful lot of precendent for conditional
dependencies, so I modeled this after the way it was done for the opensp
module.

### DIFF
--- a/devel/git/DEPENDS
+++ b/devel/git/DEPENDS
@@ -4,6 +4,12 @@ depends openssl
 depends rsync
 depends Python
 
+if [ "$(get_module_config DOCS)" == "y" ]
+then
+    depends asciidoc
+    depends xmlto
+fi
+
 optional_depends pcre \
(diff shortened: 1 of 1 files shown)


lunar-linux-b01:
SUCCESS!

v4hn:
Up to now, there just isn't *any* conditional dependencies support in lunar,
although some devs hacked something for individual modules.
See https://github.com/lunar-linux/issues for a collection of relevant features
missing from lunar. You're welcome to work on these :-)

Your case here is much more easy though.
Just remove the mquery, make asciidoc and xmlto optional_depends,
and add "also required for docs" to the second one.
`grep -r "also required" --include DEPENDS $MOONBASE`
lists some cases of this pattern in current moonbase.
Afterwards, you can replace `get_module_config DOCS`
by `in_depends $MODULE asciidoc` in BUILD.

PR #581, opened 2016-04-04
ffmpeg: version bumped to 3.0.1.

Florin65:
Some modules in the moonbase could not support the new API. for some of them there are patches for that.


lunar-linux-b01:
SUCCESS!

PR #589, opened 2016-04-19
samba: Version up to 4.4.2.

dagbrown:
This changes the build a bit, and the configuration flags, but now it
supports being an Active Directory controller, which is pretty neat.

Pull request because this is a really big version bump.


lunar-linux-b01:
SUCCESS!

PR #622, opened 2016-05-31
libcgroup: bump to 0.41. Adding a BUILD with OPTS --enable-opaque-hie…

stumbles:
…rarchy=name=systemd

so it will play nicely with systemd. Installing a few conf files as samples.


lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

PR #637, opened 2016-06-28
Sphinx: Changed the url to be https.

dagbrown:
Odds are this will need to be done for all of the other python modules
too.


lunar-linux-b01:
SUCCESS!

PR #673, opened 2016-08-30
setuptools: version bump

rtlanceroad:


lunar-linux-b01:
SUCCESS!

Florin65:
the new versions of setuptools seems to have missing some functionality and breaks some of the pythom modules. we won't update the version for setuptools until this issue is not solved.

PR #675, opened 2016-09-03
lxdm: remove non-existed lxdm.pam line

rtlanceroad:


rtlanceroad:
The error is irrelevant. `gdk-pixbuf` is a dependency for either `gtk+-2` or `gtk+-3`, which are listed as optional dependency for lxdm.

rtlanceroad:
@Florin65 Can you check this?

Florin65:
fixed in the moonbase. i actually added the missing file.

PR #685, opened 2016-09-19
emacs: version bumped to 25.1.

Florin65:
Bring more improvements to the default module.


Florin65:
rtlanceroad commented 9 hours ago
That's a known issue with Docker. https://debbugs.gnu.org/cgi/bugreport.cgi?bug=23529

Well, this might be true.

PR #686, opened 2016-09-20
clamav: Version bumped to 0.99.2.

dagbrown:
Also, this changes the URL as they seem to have left Sourceforge behind.


lunar-linux-b01:
SUCCESS!

PR #710, opened 2016-10-19
crafty: version bumped to 25.1

bsfc1685:
unzip'ing the crafty archive puts all the files in the current directory (/usr/src),
not in it's source directory (/usr/src/crafty-25.1), so I added a mv command
in the BUILD file.


bsfc1685:
argh, missing workaround sed -i '/G_DISABLE_DEPRECATED/d' ?/Makefile for gnome-vfs... I wouldn't have thought this module required those deps...

games/crafty/BUILD
```diff
@@ -1,6 +1,7 @@
 (
+mv $SOURCE_DIRECTORY/../*{.c,.h,.hlp,Makefile,speak} $SOURCE_DIRECTORY &&
```
bsfc1685: I'll check out /var/lib/lunar/functions/sources.lunar sure :)

PR #711, opened 2016-10-19
crafty: version bumped to 25.1

bsfc1685:


Ratler:
@bsfc1685 just hold on for a sec, I'm fixing gnome-vfs :)

PR #714, opened 2016-10-21
VirtualBox: remove chmod on kernel device from startup script

v4hn:
This script is supposed to be used by regular users.
They are usually not privileged to run these commands.

Also if the admin failed to setup the right permissions for these devices,
this is surely not the place to fix this.


lunar-linux-b01:
SUCCESS!

PR #717, opened 2016-10-22
angband: version bumped to 4.0.5 + changed (un)install

bsfc1685:
fixed the missing directories and wrong group in /var/games/angband which were making the game to fail to start


lunar-linux-b01:
SUCCESS!

PR #721, opened 2016-10-23
pushover: version bumped to 0.0.5

bsfc1685:


bsfc1685:
audiofile's alsa-lib dep have to be merged before

PR #736, opened 2016-11-14
VirtualBox: major version bump

v4hn:
works without problems for my setup


lunar-linux-b01:
SUCCESS!

Ratler:
@v4hn could you verify whether your VirtualBox is linked against qt4 or qt5? I find quite a few references to qt4 in the source code but it only seem to build using qt5. With qt5 it works fine for me :+1:

v4hn:
You're right. They depend on Qt5 now.
I updated the request.

lunar-linux-b01:
SUCCESS!

PR #738, opened 2016-11-18
eigen3: major version bump

v4hn:
After THREE YEARS! they finally managed to release.
Ubuntu distributed the beta for quite some time already...


lunar-linux-b01:
SUCCESS!

PR #740, opened 2016-11-21
ImageMagick: Another version bump.

dagbrown:
Also, they deleted the previous version from their downloads directory.
It's really annoying when upstream devs do that, isn't it?


lunar-linux-b01:
SUCCESS!

PR #741, opened 2016-11-24
DBD::SQLite - Version bumped to 1.52.

dagbrown:
Annoyingly, the previous version was removed from the source location.


lunar-linux-b01:
SUCCESS!

PR #747, opened 2016-11-29
fdk-aac: use github for download.

Florin65:


Florin65:
this might be a gcc 6.2.0 bug.

Ratler:
Why the switch to github? They still create a dist version without the need to regen autoconf on sourceforge.

Florin65:
yes, true. it's the same error at compile though :(

PR #759, opened 2016-12-21
cdrtools: version bumped to 3.02a07.

fbrosson:


lunar-linux-b01:
SUCCESS!

fbrosson:
Thanks @Florin65 for the merge.
BTW, I noticed that at least one contributors, @v4hn, is using the new _"Squash and merge"_ feature that  like very much and that GitHub has added earlier this year. It's very handy because it saves a _merge commit_ in the git log.
PR 755 was merged with this new _"Squash and merge"_ feature.
Of course, this only a suggestion, and I might very well be unaware of one or more reasons why traditional _Merge commits_ are still being used ;-)

v4hn:
Yeah, it depends.
As commits in the moonbase operate on isolated modules (mostly) and nobody has to cherry-pick them to other branches, it does not make much of a difference from a pragmatical point of view.

One practical difference is that you can merge requests from the command-line via `git merge` if you added the contributor's remote before.
I'm not aware of a way to do this with squash&merge. Cherry-picking works fine to *do this* from the command line, but ttbomk github does not pick these changes up as a successfully merged request.

fbrosson:
OK, thanks @v4hn for the info.

That said, I guess the [Git network graph](https://github.com/lunar-linux/moonbase-other/network) would look nicer if individual commits were merged with the _squash and merge_ feature.
And I guess that processing a repo made of such commits is faster than processing a repo with merge commits.
Well, I guess it does not make much difference with today's servers.

PR #762, opened 2016-12-23
New module: icu-le-hb 1.0.3

bsfc1685:
This module is the new support for the ICU layout engine using harfbuzz. Once installed, layoutex/libiculx.so (removed in icu4c-58) will be enabled/included when building icu4c


lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

PR #774, opened 2017-01-17
awesome: Version bump to 4.0

rtlanceroad:
Lua configuration API might be different from 3.5.9.


rtlanceroad:
ImageMagic always remove old source from their ftp, so what the hell with them....

PR #783, opened 2017-01-19
lxpanel: Version bump to 0.9.2

rtlanceroad:


rtlanceroad:
The error is caused by gtk+-2.0 as a optional dependency, it should get successfully built with gtk+-2.

PR #785, opened 2017-01-19
conky-manager: Version bump to ubuntu16.04.1

rtlanceroad:


rtlanceroad:
This bump needs `cairo` module built with pdf(`--enable-pdf`), which is provided by `ghostscript` module. But for `cairo`, `ghostscript` is optional. _sigh_

PR #787, opened 2017-01-21
light-locker: Version bump to 1.6.0

rtlanceroad:


rtlanceroad:
The error msg:
```
checking for cairo-pdf.h... no
configure: error: 
*** Can't find cairo-pdf.h. You must build Cairo with the pdf
*** backend enabled.
```
This needs `cairo` built with `ghostscript`.

PR #788, opened 2017-01-21
screen-message: Version bump to 0.24

rtlanceroad:


rtlanceroad:
The error msg:
```
checking for cairo-pdf.h... no
configure: error: 
*** Can't find cairo-pdf.h. You must build Cairo with the pdf
*** backend enabled.
Creating /var/log/lunar/compile/gtk+-2-24.31.xz 
! Problem detected during BUILD
```
This still needs `cairo` built with `ghostscript`, which is a optional dependency.

PR #793, opened 2017-01-22
pulseaudio: optional udev defaults to n

bsfc1685:
I want to see if a pulseaudio's configure doesn't fail with this
(later comment edits: catched while doing an update after rtlaneroad's first commit on i3status PR) ... we talked on IRC about it and to let florin have a thought on that
it was closed because I saw "pending" on his PR and thus wanted to let rtlanceroad have another i3status build, but I saw jenkins can trigger more than one at once :P

### DIFF
--- a/audio/pulseaudio/DEPENDS
+++ b/audio/pulseaudio/DEPENDS
@@ -3,6 +3,8 @@ depends libsndfile
 depends libatomic_ops
 depends gdbm
 depends speexdsp
+depends json-c
+depends sbc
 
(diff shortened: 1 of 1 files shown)


lunar-linux-b01:
SUCCESS!

bsfc1685:
final check

lunar-linux-b01:
SUCCESS!

bsfc1685:
weird... when udev defaults to yes, it's not actually built/installed on the jenkins host... at least not configure-detected

PR #803, opened 2017-01-28
cairo: Added poppler (PDF) dependency required by libreoffice-bin

bsfc1685:
then the PDF support will be included in cairo as reclaimed by libreoffice-bin compile error on /usr/include/cairo-pdf.h


lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

PR #804, opened 2017-01-29
cairo: changed cairo's poppler dep to required

bsfc1685:


lunar-linux-b01:
SUCCESS!

bsfc1685:
well it's "--enable-pdf=[no/auto/yes]" as per ./configure --help |grep enable-pdf ...  and I guess --disable/enable-pdf by the configure run

lunar-linux-b01:
SUCCESS!

bsfc1685:
the build succeeded, but the PDF support still is not enabled... I don't get it :P libreoffice-bin will wait unless someone else has an idea, somewhen, sorry there :)

PR #808, opened 2017-01-30
icu4c: Updated broken URL

bsfc1685:


lunar-linux-b01:
SUCCESS!

v4hn:
The current URL redirects to sourceforge mirrors already. Because it's the one provided by upstream, I would like to keep it as is.
If you really want to replace it with the direct sourceforge link, please use the `$SFORGE_URL` variable instead, so you can select your mirror in your lunar config.

PR #810, opened 2017-01-30
gtk+-3: Added poppler as required dep

bsfc1685:


bsfc1685:
I've had enough :|

v4hn:
@bsfc1685 I see why, you've been fighting with this for a while now..

@Ratler looks like this failed because the server didn't rebuild cairo for the pull-request and the cairo installation used for testing the request does not have pdf support enabled (the default for this dep changed in #810 ). Could you have a look please?

PR #884, opened 2017-10-19
gdk-pixbuf: Fix up some dependencies.

dagbrown:
Now it depends on shared-mime-info.  And because apparently the
gdk-pixbuf developers haven't been testing how well their package honors
the configuration flags you request, tiff is now a hard dependency
rather than an optional one.


lunar-linux-b01:
SUCCESS!

PR #885, opened 2017-10-19
cairo: Break the dependency loop.

dagbrown:
When you install librsvg (which is a stricly optional dependency; cairo
builds A-OK without it)), it gives you the option to rebuild cairo with
SVG support anyway.  So let's get rid of it here and lose the circular
dependency.


lunar-linux-b01:
SUCCESS!

PR #887, opened 2017-10-25
libostree: New module, for managing binary deployment and upgrades.

dagbrown:
In library form, rather than as an OS management tool, it's also used by
things like "flatpak".


lunar-linux-b01:
SUCCESS!

PR #906, opened 2017-11-30
curl:Update download URL

lhw828:
curl:Update download URL
http method can not download the software.


lunar-linux-b01:
SUCCESS!

PR #907, opened 2017-12-01
Samba fix startup

dagbrown:
Samba has moved its daemons into /usr/sbin, which is correct!  But our systemd units still have it in /usr/bin which means that it fails now.


lunar-linux-b01:
SUCCESS!

PR #909, opened 2017-12-02
nettle:Update Download URL

lhw828:
The original link is invalid and cannot be installed normally after downloading,deletes an invalid link.


lunar-linux-b01:
SUCCESS!

stumbles:
The http://www.lysator.liu.se/~nisse/archive works fine here. Perhaps it was an issue with your ISP or a temporary issue with their site. 

In any case, perhaps delete your PR and resubmit with ftp.gnu.org/gnu/nettle/ as a third source url.

PR #912, opened 2017-12-04
x11-utils:Update some download links to HTTPS

lhw828:
Update some download links to HTTPS


lhw828:
Ask the administrator to manually check this push, the reason is not compiled compilation vala tips "No package 'libgvc' found", rather than the problem I push.

PR #933, opened 2017-12-11
chromium-arch: chromium binary builded from Arch Linux.

Florin65:


lunar-linux-b01:
SUCCESS!

lhw828:
@Florin65 Thank you！

PR #938, opened 2017-12-14
libquicktime: adding another patch but keeping one around for now. As…

stumbles:
… is compiles and appears to be fine but you never know


lunar-linux-b01:
SUCCESS!

PR #942, opened 2017-12-15
rsyslog: adding a mkdir to BUILD so its service file stops barking ab…

stumbles:
…out /var/spool/rsyslog missing


lunar-linux-b01:
SUCCESS!

PR #952, opened 2017-12-20
httpd: Version bump to 2.2.34.

dagbrown:
Note that with this release, httpd 2.2 is now at its end-of-life, so if
you haven't done so already, you should migrate to httpd 2.4.


lunar-linux-b01:
SUCCESS!

PR #954, opened 2017-12-21
Fixed the URL for bitcoin-core.

dagbrown:
As it says on their github page:

> Preferably use the above download link, not the below links to
> download the source tarball, as the release tarballs are generated
> deterministically and GitHub's are not.


lunar-linux-b01:
SUCCESS!

PR #959, opened 2017-12-22
w3c-libwww: Adding a sedit to delete a case dealing with SSLv3 else t…

stumbles:
…he build breaks

with no reference to it.


lunar-linux-b01:
SUCCESS!

PR #971, opened 2017-12-25
Zfs utils dracut fixes

dagbrown:
This fixes a couple of issues with dracut's zfs module.  One is a problem on our end (well, not a problem, just a thing we do differently), I think, and another is a problem on the ZFS end.


lunar-linux-b01:
SUCCESS!

PR #978, opened 2017-12-28
volume_key: Need to add a patch due to recent cryptsetup version bump…

stumbles:
…. Shamelessly borrowed from the Arch Linux folks.

Thanks guys.


lunar-linux-b01:
SUCCESS!

PR #986, opened 2017-12-31
distcc: The source url moved and they apparently made some other chan…

stumbles:
…ges at that time

since BUILD needed a fix.


lunar-linux-b01:
SUCCESS!

PR #1010, opened 2018-01-06
mpv: made libcdio-paranoia a required dependency

bulletfreak:
it refuses to build without it


stumbles:
You need to put --disable-libass in BUILD, or add libass as an optional_depends.

Florin65:
I have added the libass optional depend as stumbles proposed.
The module build without libcdio-paranoia installed.

PR #1017, opened 2018-01-13
serd: Adding to moonbase. Part of a dependency chain required for a t…

stumbles:
…raverso bump which now requires

lilv.


lunar-linux-b01:
SUCCESS!

PR #1025, opened 2018-01-16
gtk+-2: Somebody with NVIDA please check this. Just compiled it here …

stumbles:
…with that stuff

disabled and it went fine. I need a sanity check please. This is with NVIDIA 390.12 so I am curious
if the same happens with current moonbase version 384.98.


Ratler:
@stumbles I just merged NVIDIA 390.12, and this change works just fine with it.

stumbles:
Good then Ratler. I will commit this.

stumbles:
I will leave the comments for future reference.

PR #1042, opened 2018-02-15
qemu: Another patch, to fix compiling with gcc 7.x.

dagbrown:
In newer gcc's (starting at 7.2?), cpp exits with fatal status if you
give it the (meaningless) "-c" flag (I'm assuming that's the flag you
normally give to the compiler to tell it not to link after compiling).


lunar-linux-b01:
SUCCESS!

Florin65:
The version was bumped and the module builds without this patch.

PR #1050, opened 2018-03-17
lynx: Added ncurses patch to fix bug

bulletfreak:
https://lists.gnu.org/archive/html/bug-ncurses/2017-03/msg00015.html

It refuses to build without the patch giving the error linked above


lunar-linux-b01:
SUCCESS!

PR #1055, opened 2018-03-28
icu4c: version bumped to 61.1

Florin65:
This will trigger at 'lunar fix' a lot of rebuilds in the system.
Treat it with carefull!


lunar-linux-b01:
SUCCESS!

PR #1059, opened 2018-04-04
Beep patches

dagbrown:
Here are a couple of patches from the guys at Debien:

1) catch-sig-term.patch: If someone bangs on ^C halfway through a beep, make the PC speaker shut up, and

2) CVS-2018-0492.patch: The beep command had a security problem!  Imagine that being how your system got rooted.  How embarrassing would that be?


lunar-linux-b01:
SUCCESS!

PR #1106, opened 2018-06-28
pkgconfig: Move from moonbase-core to moonbase-other.

dagbrown:


lunar-linux-b01:
SUCCESS!

v4hn:
It's somewhat frustrating that someone notices this only now, but we currently have `pkgconfig` in both `moonbase-core` and `moonbase-other`.

@dagbrown created this request together with https://github.com/lunar-linux/moonbase-core/pull/1913 , which is not merged yet because it contains other changes and hardcodes pkgconf everywhere for some reason.

Personally, I didn't see a reason to move to `pkgconf` in the first place, but I guess it makes sense for you.

@dagbrown @Florin65 can we please resolve this and remove one of the two packages?

Issue #1134, opened 2018-08-17
boost:pyconfig.h not found.


lhw828:
boost:pyconfig.h not found.

https://i.loli.net/2018/08/17/5b76307890d12.png

stumbles:
Corrected with commit f6d96ecaa09cc926dd29df5f761315c855a81cce.

PR #1136, opened 2018-08-17
boost:Change python's default options.

lhw828:
boost:Change python's default options to n.


lhw828:
https://github.com/lunar-linux/moonbase-other/issues/1134

PR #1142, opened 2018-08-19
poppler: Update to 0.68.0 & make openjpeg-2 and `%JPEG` regular deps

stratact:
openjpeg-2 and `%JPEG` have to be regular dependencies, otherwise the build breaks without them.


lunar-linux-b01:
SUCCESS!

PR #1143, opened 2018-08-21
Move nonessential filesystems

dagbrown:
Time to clean up some clutter in moonbase-core: move jfs and ntfs to moonbase-other where they beling.


lunar-linux-b01:
SUCCESS!

PR #1157, opened 2018-09-03
fuse:Change download URL to https

lhw828:
fuse:Change download URL to https
And the releases is 3.2.6 now.
https://github.com/libfuse/libfuse/releases


lunar-linux-b01:
SUCCESS!

PR #1169, opened 2018-09-15
exiv2: fixed patch shasum

Duskmourn:


Florin65:
Than you for notification. I have fix this manually.

PR #1199, opened 2018-10-29
resolvconf: Update

Duskmourn:


Florin65:
The %MODULE is a typo. It have to be $MODULE.

Florin65:
Will have to find a fix for:
   mv: cannot move '/etc/resolv.conf' to '/etc/resolvconf/resolv.conf.d/original': Device or resource busy

Florin65:
The module is not needed anymore in the moonbase.

v4hn:
> The module is not needed anymore in the moonbase.

That's a bit harsh to say, but systemd-resolved provides a functionally superior alternative.

PR #1205, opened 2018-11-09
xapian-core:Change download URL to https

lhw828:
xapian-core:Change download URL to https and add web site.


lunar-linux-b01:
SUCCESS!

PR #1216, opened 2018-11-26
 polkit:version bumped to 0.115 

lhw828:
polkit:version bumped to 0.115 、add a depends and  remove the patch.


lunar-linux-b01:
SUCCESS!

Florin65:
Bumped manually the version to 0.115. Thank you for this. The mozjs depend is not needed (spidermonkey is the js engine needed).

lhw828:
Stumbles and I came to the same conclusion that mozjs was needed.
He has proposed a new PR.

PR #1249, opened 2018-12-27
keepassxc: add to moonbase

Duskmourn:


florin65:
Added a similar modules in the moonbase.

PR #1282, opened 2019-02-14
adding osinfo helper packages into moonbase

lazyape:


lazyape:
@florin65 please have a look here

PR #1296, opened 2019-02-20
libcanberra:Dependency libcanberra-gtk3

lhw828:
libcanberra:gnome-bluetooth Dependency libcanberra-gtk3

### DIFF
--- a/audio/libcanberra/DEPENDS
+++ b/audio/libcanberra/DEPENDS
@@ -1,5 +1,6 @@
 depends libtool
 depends libvorbis
+depends gtk+-3
 
 optional_depends %UDEV "--enable-udev" "--disable-udev" "for udev support"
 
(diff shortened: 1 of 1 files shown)


Florin65:
There is no need for this. libcanberra is a depend for gnome-bluetooth.
What's needed is that the libcanberra module to be installed after/with  GTK+-3.

lunar-linux-b01:
SUCCESS!

lazyape:
> 
> 
> There is no need for this. libcanberra is a depend for gnome-bluetooth.
> What's needed is that the libcanberra module to be installed after/with GTK+-3.

not even that. we need to ensure that libcanberra has being built with gtk+-3 enabled
something like

if ! in_depends libcanberra gtk3; then
  echo "Recompile libcanberra with gtk+-3 support"
  false
fi &&

PR #1297, opened 2019-02-25
bind: Fixing module name.

dagbrown:
We seriously need to set moonbase-other to be PR-only.  The number of
terrible, broken commits here lately is just unacceptable.


lunar-linux-b01:
SUCCESS!

PR #1311, opened 2019-03-18
polkit: disable documentation buildind

lazyape:
i hope that this will prevent xsltproc error on jenkins


lunar-linux-b01:
SUCCESS!

PR #1312, opened 2019-03-26
freetype2: Version bumped to 2.10.0

dagbrown:
The patches completely failed to apply, so I'm going to just to assume
that the problems that the patches sorted out were fixed within
freetype2 itself.


lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

PR #1325, opened 2019-05-14
subversion: Bump to version 1.12.0.

dagbrown:
Also add in a patch that fixes building with APR 1.7.0.


lunar-linux-b01:
SUCCESS!

PR #1348, opened 2019-06-24
chrono: version bumped to 0.2.27

cody1632:
Now process months and years. 
Also fixed a bug in the countdown parsing code.
Rewrote time string display function.


lunar-linux-b01:
SUCCESS!

PR #1349, opened 2019-06-26
chrono: version bumped to 0.2.29

cody1632:
-Now reset even when paused.
-Count seconds_total according to months of 28/29/30/31 days.
-Handle leap year
-Added tests for years and months parsing


lunar-linux-b01:
SUCCESS!

PR #1362, opened 2019-08-06
create-headers: Version bump to 5.1.10.

dagbrown:
This goes along with the kernel-headers bump.


florin65:
Too old PR. Will have to be renewed.

PR #1406, opened 2020-02-29
spice-gtk: Unified vala and introspection in one optional dependency

lazyape:
Both are needed from virtual management tools like virt-manager


florin65:
It's better to keep them separate though. We can put "sane-defaults" though. I was managed years ago to install only gtk-based modules without gobject-introspection. Some might want install gobject-introspection without vala (no libvirt need).

PR #1414, opened 2020-03-11
 texstudio-qt5:fix a depends typo

lhw828:
texstudio-qt5:fix a depends typo


v4hn:
The problem was actually a whitespace after the `\` that should have escaped the newline. [I committed a fix](https://github.com/lunar-linux/moonbase-other/commit/c013d367a8d07e3b4aee8726df8b171b57c618fa).

PR #1415, opened 2020-03-11
chromium-arch:fix hash

lhw828:
chromium-arch:fix hash


v4hn:
Thanks for cleaning up...

PR #1429, opened 2020-04-03
openttd: version bumped to 1.10.0

cody1632:


cody1632:
well I don't know why racket is triggered and fails, but it will have to be fixed first...

PR #1435, opened 2020-04-27
pinentry: added warning about circular dependency

tbeckerson:
pinentry has an optional_depends on gcr.
gcr requires gnupg.
gnupg requires pinentry.

pinentry -> gcr -> gnupg -> pinentry -> gcr -> etc

This PR adds a warning to the optional_depends, and defaults to choosing no.


lunar-linux-b01:
SUCCESS!

PR #1442, opened 2020-05-05
Inkscape: version bump

tbeckerson:
Points of interest:

BUILD: 
- libcdr doesn't exist in Lunar, disabled.
- GraphicsMagick is based on ImageMagick-7, which Inkscape does not support. Explicitly disabled.
DEPENDS:
- Almost entirely new, using GTK3.
DETAILS:
-Version bump, update sha256.
-*.patch: Was for the previous version of Inkscape, and created compatibility for ImageMagick v6, which Lunar doesn't have anyway. Removed entirely.


lunar-linux-b01:
SUCCESS!

PR #1455, opened 2020-06-01
scour: SVG Optimizer

tbeckerson:
Scour is a commandline SVG optimizer.

BUILD: scour doesn't have a ./configure file, this skips straight to default_make.
The rest is standard module


lunar-linux-b01:
SUCCESS!

PR #1456, opened 2020-06-02
geeqie:version bump

tbeckerson:
BUILD: the install comes with a configure file now, changed to standard default_build.
DEPENDS: updated to reflect all available options.
DETAILS: update version, sha256, and modification date.

PRE_BUILD & patch file are no longer necessary to compile.


tbeckerson:
according to the log at https://serenity.stderr.eu/ci-lunar/job/moonbase-other/1377/consoleText , this fails with:
'fatal error: gtk/gtk.h: No such file or directory'
But, it also claims that gtk+-3 is pre-installed, which should include gtk.h.
I'm not sure how to proceed from here.

PR #1463, opened 2020-06-25
intel-tbb: fix broken install for 32bit systems

cody1632:
Small change on the output dir which starts with "linux_intel*" on 64bit systems but is generated as "linux_ia32*" on 32bit systems.


lunar-linux-b01:
SUCCESS!

PR #1500, opened 2020-09-22
Argyll - version bump, gcc 10 fix

tbeckerson:
- Updated DETAILS to newest version, 
- Added patchfile to work with GCC 10


lunar-linux-b01:
SUCCESS!

PR #1506, opened 2020-09-26
libmspub - GCC 10 fix

tbeckerson:
patch to work with GCC 10. Libreoffice-bin needs this by default.


florin65:
Retest this please

florin65:
Retest this please

PR #1508, opened 2020-09-29
httpd: fix mpm warnings in logs

cody1632:


cody1632:
This fix removes warnings from httpd error_log:

> [Tue Sep 29 04:54:17.700132 2020] [http2:warn] [pid 341] AH10034: The mpm module (prefork.c) is not supported by mod_http2. The mpm determines how things are processed in your server. HTTP/2 has more demands in this regard and the currently selected mpm will just not do. This is an advisory warning. Your server will continue to work, but the HTTP/2 protocol will be inactive.

Looks like the default mpm is prefork which offers lesser performances.

lunar-linux-b01:
SUCCESS!

PR #1510, opened 2020-10-05
rsync: add optional depends

cody1632:


cody1632:
On freshly installed Lunar, rsync fails to configure due to missing lz4.

lunar-linux-b01:
SUCCESS!

PR #1513, opened 2020-10-22
mozjs78 - new version

tbeckerson:
Gnome is already using mozjs 78 - gjs fails without it.

This is almost unchanged from the mozjs already found in the moonbase.


lunar-linux-b01:
SUCCESS!

PR #1517, opened 2020-11-21
libhandy: version bump, it is an optional dependency for gnome-control-center

lazyape:


lunar-linux-b01:
SUCCESS!

lazyape:
@florin, i propose to move libhandy under gnome3/libs

PR #1526, opened 2020-12-11
geeqie - version bump

tbeckerson:
- Version bump to 1.6
- There's a problem with gpu acceleration (clutter-gtk -> libX11). It's a known issue, and will hopefully be patched soon.
See: https://github.com/BestImageViewer/geeqie/issues/829, which leads to
https://gitlab.freedesktop.org/xorg/lib/libx11/-/issues/125


lunar-linux-b01:
SUCCESS!

PR #1527, opened 2020-12-13
gdb: Add source-highlight as an optional dependency

lazyape:
Gdb supports source highlighting of the debugged program throught source-highlight module.
This behavior can be controlled at runtime by set style sources ‘on|off’ in the gdb console or the .gdbinit file


lunar-linux-b01:
SUCCESS!

PR #1531, opened 2020-12-23
gmic - image processing framework

tbeckerson:
This is just the libraries and CLI. If/when this is merged, I'll add _gmic-qt,_ which contains the GUIs.


lunar-linux-b01:
SUCCESS!

PR #1533, opened 2020-12-27
libopenraw - 0.3.0 updates

tbeckerson:
tweaks to the recent 0.3.0 update

- BUILD: the sedit breaks the build if it's used (error: 'JPEG' has not been declared), and has been removed.
- DEPENDS: moved gdk-pixbuf to optional_depends, and added rustc option.
- DETAILS: I'm still using the .xz instead of the .bz2. Let's save upstream a 100K of bandwidth.
- PRE_BUILD: fixes an error with the openraw-gnome pkgconfig file. This has been merged upstream.


lunar-linux-b01:
SUCCESS!

PR #1536, opened 2020-12-27
libraqm - library and API for text layout

tbeckerson:
new module!

The raqm libraries are supported in ImageMagick and libgd (I'll update those if this is merged).


lunar-linux-b01:
SUCCESS!

PR #1537, opened 2021-01-09
geeqie - patch to fix clutter render

tbeckerson:
The 1.6 update had a known bug where gpu acceleration (using clutter) was causing segfaults. This PR includes a patch that fixes the problem, and restores clutter as a DEPENDS option.


lunar-linux-b01:
SUCCESS!

PR #1538, opened 2021-01-09
Graphics magick

tbeckerson:
- version bump to 1.3.36
- removed some un-necessary depends (bzip and xz are already installed by default)


lunar-linux-b01:
SUCCESS!

PR #1544, opened 2021-01-29
node - use Python3

tbeckerson:
- Node uses Python v3 by default, updated DEPENDS.
- Also removed unused dependency on 'v8'


lunar-linux-b01:
SUCCESS!

PR #1548, opened 2021-02-12
chromium: migrate to bundled libraries & bump to a recent dev version

v4hn:
DISCLAIMER:
Lunar's current chromium module was unmaintained and broken for a long time (chromium-arch provided a hacky replacement to use arch-linux's build).
So you can judge these changes any way you like, but at least they should provide a working chromium binary again.

---

Lunar does not have the resources to fix chromium's build with system libraries every single time something breaks.
So let's migrate to a simpler build that is better supported upstream.

This means we let them compile with their own bundled libs so the module does not break with every minor API change in newer releases.
The "official build" would also have to rely on provided binaries *including a clang compiler*, but that would be just ridiculous.

90.0.4407.0 is an arbitrary nightly dev version. Feel free to change this build to any other (more offical) version.
It just that I picked this as the latest version when I got started and it takes forever to fix issues and build this module.
So I will not play with it anymore until a more current official release comes out as a beta or stable version.


esselfe:
Closing, this PR is broken and old.

PR #1560, opened 2021-04-13
darktable update

tbeckerson:
Whole bunch of changes - the darktable module hasn't been updated in over a year.

BUILD - darktable either requires gcc to be compiled with isl support, or clang to be used as a compiler instead. Since clang will already be installed as part of the dependency chain (gtk+-3 > mesa-lib > llvm), it's easier & safer to use clang.

DEPENDS - this has been almost entirely rewritten because of changes between v2.6 and v3.4

DETAILS - Normal version bump changes.

PRE_BUILD - darktable only checks up to v11.0 of llvm, I've added a sed line to include v11.1.

### DIFF
--- a/graphics/darktable/BUILD
+++ b/graphics/darktable/BUILD
@@ -1,5 +1,4 @@
-# GnomeKeyring is only supported for versions older than 3.12.0, version 3.12.0 found.
-# Please use libsecret instead.
-OPTS+=" -DBUILD_USERMANUAL=0 -DDONT_USE_INTERNAL_LUA=0" &&
-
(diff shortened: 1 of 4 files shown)


tbeckerson:
Compile broke because of errors earlier in the dependency chain (libpsl). I'll try to fix that first.

florin65:
Retest this please

lunar-linux-b01:
SUCCESS!

v4hn:
A lot happened there for sure. I pushed a fixup commit to simplify some things and CI succeed now.

Thanks @tbeckerson, great to have this updated!

PR #1562, opened 2021-04-13
libpsl + publicsuffix-list fixes

tbeckerson:
libpsl:
DEPENDS - fixed an error in the gtk-doc optional_depends that causes a configure failure if you chose to _not_ install (libpsl defaults to enabling docs).
PRE_BUILD - autoreconf doesn't read the OPTS in the DEPENDS file (because PRE_BUILD is finished before DEPENDS is read), which meant the gtk-doc failure still happened. It's not necessary to the build anyway.
BUILD - using default_build works properly, even with the exported LC_CTYPE (why aren't we using the system default?). This makes the autoreconf in PRE_BUILD unnecessary.
POST_INSTALL - since we're deleting the suffix list in PRE_BUILD, I added in this implicit install of _publicsuffix-list,_ which is the whole point to libpsl.

publicsuffix-list:
DETAILS: updated to the most recent commit.

### DIFF
--- a/libs/libpsl/BUILD
+++ b/libs/libpsl/BUILD
@@ -5,8 +5,7 @@ OPTS+=" --disable-static \
         --enable-runtime=libidn2 \
         --with-psl-file=/usr/share/publicsuffix/effective_tld_names.dat \
         --with-psl-testfile=/usr/share/publicsuffix/test_psl.txt"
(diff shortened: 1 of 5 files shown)


lunar-linux-b01:
SUCCESS!

PR #1563, opened 2021-04-14
deadbeef - version bump

tbeckerson:
BUILD:
- gtk2 is now legacy code, and upstream will stop supporting it soon.
- last.fm and notification support require swift/libdispatch, which we don't have on Lunar.

DEPENDS:
- made gtk3 the default
- fixed an error in the faad2 optional_depends

DETAILS:
normal version bump updates

PRE_BUILD:
there's a string replacement that doesn't seem to be necessary anymore (compiled and works fine without it), but I've commented it out and left it here in case I'm wrong and it is needed.


lunar-linux-b01:
SUCCESS!

PR #1564, opened 2021-04-23
qrencode - update DEPENDS

tbeckerson:
_qrencode_ installs using SDL2 without problems. I think we should use the newer library when possible.


lunar-linux-b01:
SUCCESS!

PR #1566, opened 2021-05-14
wireshark: fix DEPENDS for cmake

esselfe:


net/wireshark/DEPENDS
```diff
@@ -1,30 +1,47 @@
...
-optional_depends gtk+-3 \
-                 "--enable-wireshark --with-gtk=3" "" \
-                 "for GTK 3.x based frontend (select this or the next one gtk+-2, not both)"
-
-optional_depends gtk+-2 \
-                "--enable-wireshark --with-gtk=2" "" \
-		"for GTK 2.x based frontend (select this or the previous one gtk+-3, not both)"
+depends libxml2
+depends perl
+depends qt5
```
lazyape, line 9: this prevent build in headless systems. qt5 should be a dependency only if gui is selected (BUILD_wireshark in CMakeOptions.txt)

esselfe: Oh boy, didn't see that file! I was only looking in CMakeLists.txt...
Thanks

lunar-linux-b01:
SUCCESS!

PR #1569, opened 2021-05-29
graphite - update DEPENDS

tbeckerson:
Graphite no longer needs python2, and probably has not for a while. This was probably an old DEPENDS that was never updated because it never actually failed to build.


lunar-linux-b01:
SUCCESS!

PR #1570, opened 2021-05-31
new module - python-appdirs

tbeckerson:
A small Python module for determining appropriate platform-specific dirs, e.g. a "user data dir"

This module is for multi-platform apps to figure out the correct directory structure to use when storing user data.
From the homepage:

> What directory should your app use for storing user data? If running on Mac OS X, you should use:
> 
> ~/Library/Application Support/<AppName>
> 
> If on Windows (at least English Win XP) that should be:
> 
> C:\Documents and Settings\<User>\Application Data\Local Settings\<AppAuthor>\<AppName>
> 
> or possibly:
> 
> C:\Documents and Settings\<User>\Application Data\<AppAuthor>\<AppName>
> 
> for roaming profiles but that is another story.
> 
> On Linux (and other Unices) the dir, according to the XDG spec, is:
> 
> ~/.local/share/<AppName>

This module is needed for the new extensions manager in Inkscape (updated PR for that pending approval of this).

### DIFF
--- a/python/python-appdirs/DEPENDS
+++ b/python/python-appdirs/DEPENDS
@@ -1,0 +1,1 @@
+depends python-setuptools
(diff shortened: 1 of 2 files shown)


lunar-linux-b01:
SUCCESS!

PR #1593, opened 2021-08-09
syslog-ng: add systemd service

esselfe:
The service file was packaged with the source.

### DIFF
--- a/systemd.d/syslog-ng.service
+++ b/systemd.d/syslog-ng.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=System Logger Daemon
+Documentation=man:syslog-ng(8)
+Wants=network.target network-online.target
+After=network.target network-online.target
+
+[Service]
+Type=notify
+ExecStart=/usr/sbin/syslog-ng -F $SYSLOGNG_OPTS -p /var/run/syslogd.pid
+ExecReload=/bin/kill -HUP $MAINPID
+EnvironmentFile=-/etc/default/syslog-ng
+EnvironmentFile=-/etc/sysconfig/syslog-ng
+StandardOutput=journal
+StandardError=journal
+Restart=on-failure
+
+[Install]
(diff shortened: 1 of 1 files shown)


systemd.d/syslog-ng.service
```diff
@@ -0,0 +1,18 @@
...
+Description=System Logger Daemon
+Documentation=man:syslog-ng(8)
+Wants=network.target network-online.target
+After=network.target network-online.target
+
+[Service]
+Type=notify
+ExecStart=/usr/sbin/syslog-ng -F $SYSLOGNG_OPTS -p /var/run/syslogd.pid
+ExecReload=/bin/kill -HUP $MAINPID
+EnvironmentFile=-/etc/default/syslog-ng
```
Ratler, line 11: This line can be removed, we don't use this in lunar.

systemd.d/syslog-ng.service
```diff
@@ -0,0 +1,18 @@
...
+EnvironmentFile=-/etc/sysconfig/syslog-ng
```
Ratler, line 12: This line should point to /etc/config.d/syslog-ng (lunar does not use sysconfig).

systemd.d/syslog-ng.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=System Logger Daemon
+Documentation=man:syslog-ng(8)
+Wants=network.target network-online.target
+After=network.target network-online.target
+
+[Service]
+Type=notify
+ExecStart=/usr/sbin/syslog-ng -F $SYSLOGNG_OPTS -p /var/run/syslogd.pid
```
Ratler, line 9: Just nit-picking, but change /var/run/syslogd.pid to /run/syslogd.pid

Ratler (changes requested):
Just needs a couple of small tweaks, otherwise, it looks good.

PR #1600, opened 2021-09-29
updated depends & build

tbeckerson:
BUILD: _--enable-symbol-prefix_ causes several of the programs to fail if ImageMagick is also installed. Static files were disabled, and X support is implicitly included.
DEPENDS: added some missing deps, and GM requires ghostscript now.


lunar-linux-b01:
SUCCESS!

PR #1613, opened 2021-12-04
pulseaudio: fix missing alsa lib

esselfe:
The error I was getting was:
`E: [pulseaudio] ltdl-bind-now.c: Failed to open module module-alsa-card.so: module-alsa-card.so: cannot open shared object file: No such file or directory
E: [pulseaudio] module.c: Failed to open module "module-alsa-card".`


esselfe:
Just noticed the alsa=enabled option can be set with the "alsa-oss" option in DEPENDS :P

PR #1614, opened 2021-12-26
opencv: version bumped to 4.5.5.

florin65:


stumbles:
Jekins looks to have a problem with jasper, builds fine here. Also 4.5.5 doesn't need the lapack patch.

Closing with commit 0d4e2dd33096124fb6432edbfef993999a0d1657 bump and patch removal.

PR #1623, opened 2022-01-26
policykit: Add patch to close CVE-2021-4034 security hole

dagbrown:
Local users can become root without much difficulty (but this patch
fixes that).


lunar-linux-b01:
SUCCESS!

PR #1632, opened 2022-02-08
librsvg: fix depends

tbeckerson:
CONFIGURE: the _miscellaneous tools_ option doesn't exist anymore

DEPENDS:
 - pango requires cairo on Lunar by default
 - gtk-doc is optional
 - gobject-introspection had a typo (both _yes_ and _no_ were set to _enable_)
 - made the disable options for both _vala_ and _introspection_ match what's in the ./configure file.

### DIFF
--- a/graphics/librsvg/DEPENDS
+++ b/graphics/librsvg/DEPENDS
@@ -9,7 +9,6 @@ optional_depends vala \
                  "--enable-vala=no" \
                  "for vala support"
                  
-                 
 optional_depends gobject-introspection \
                  "--enable-introspection" \
                  "--enable-introspection=no" \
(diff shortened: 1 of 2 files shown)


lunar-linux-b01:
SUCCESS!

graphics/librsvg/DEPENDS
```diff
@@ -1,16 +1,16 @@
...
-depends pango      # needs pango with cairo enabled
+depends pango
 depends rustc
 depends intltool
 depends gtk-doc
 
 optional_depends vala \
                  "--enable-vala" \
-                 "--disable-vala" \
+                 "--enable-vala=no" \
```
v4hn, line 9: this is a standard autotools configure script, so `--disable-*` and `--enable-*=no` are equivalent. No reason to change this...

v4hn (approved):

PR #1634, opened 2022-02-09
firefox: version bump and lots of other fixes

tbeckerson:
Lots of changes!

BUILD:
 - the change to LDFLAGS and specifying SHELL aren't necessary any more, and the build works fine without them
 - got rid of the /usr/lib/lunar/plugins symlink. We don't actually do anything with this.

DEPENDS:
 - removed some dependencies that are implicitly installed by other deps. Also, good by to gtk2 and Flash!
 - since our preference is to use system-installed programs, I removed the optional_depends, made them regular depends, and added their OPTS directly into mozconfig

DETAILS:
 - normal version bump stuff
 - removed note that says our version of FF uses gtk2 and recommends to use seamonkey for full features - it hasn't used gtk2 in years, and our seamonkey module is 8 years old.
  
PRE_BUILD:
 - almost all the _export_ options work by default , or are covered by mozconfig options, and aren't necessary to the build. a couple got moved into _BUILD_ and _mozconfig._
 - MOZ_OBJDIR (build directory) is now handled by MOZBUILD_STATE_PATH.
 - rather than post a notice about using the google_api_key, then echoing the key into a file, I put a more concise notice and put the key directly into a file.
 - the problems fixed by the _sedit_ lines don't exist any more, so they've been removed.

google_api_key:
 - new file containing the Lunar api key.
 
mozconfig:
 - rearranged the options into logical groups.
 - removed the options that we had commented out for years (startup cache and cairo).
 - added in an option about webassembly sandboxing (which needs to be fixed with another module).
 - removed the explanation about disabling telemetry: no one is looking at this file, and the telemetry feature doesn't gather this info any more (according to Mozilla).

### DIFF
--- a/web/firefox/BUILD
+++ b/web/firefox/BUILD
@@ -1,14 +1,10 @@
-# this breaks a symbol check in the build scripts
-LDFLAGS=$(echo $LDFLAGS | sed "s/-s//") &&
-
-SHELL="${SHELL:-/bin/bash}" &&
-
-CFLAGS="${CFLAGS/-fno-plt/}" &&
-CXXFLAGS="${CXXFLAGS/-fno-plt/}" &&
+CFLAGS="${CFLAGS/-fno-plt/}"
+CXXFLAGS="${CXXFLAGS/-fno-plt/}"
+CC="clang"
+CXX="clang++"
 
 # LTO needs more open files
-ulimit -n 4096
-
+  ulimit -n 4096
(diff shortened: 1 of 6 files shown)


tbeckerson:
Retest this please

tbeckerson:
Retest this please

lunar-linux-b01:
SUCCESS!

PR #1647, opened 2022-02-22
php: fix wrong config path

esselfe:
HTML files starting with <?xml fail to display, giving this error message:  
"Parse error: syntax error, unexpected 'version' (T_STRING) in /srv/files/docs/tmp/index.html on line 1"  

The short_open_tag is set to Off in /etc/php.ini but phpinfo() reveals that it's set to On.

Moving php.ini in /etc/php fixes this issue and the page displays correctly.


stumbles:
Retest this please.

PR #1649, opened 2022-02-25
bitcoin: version bumped to 22.0

esselfe:


lunar-linux-b01:
SUCCESS!

esselfe:
This builds an experimental version, not recommended for production use. I didn't find a way to turn this off without the build to fail.  

After rethinking about it, I'm not too sure it's a good idea.  
Now I'll let the devs handle that instead :)

PR #1736, opened 2022-03-27
qhull: Adjusting BUILD to silence this warning during an octave compile;

stumbles:
WARNING: Qhull library not found.  This will result in loss of functionality for some geometry functions


lunar-linux-b01:
SUCCESS!

PR #1739, opened 2022-03-27
netcdf: Some adjustments to BUILD and DEPENDS.

stumbles:


stumbles:
Hmm. Well that won't work since gcc build defaults to no gfortran.

PR #1846, opened 2022-05-16
php: version bumped to 7.4.29

esselfe:
POST_INSTALL was missing a 'mkdir /etc/php' on initial installation.


esselfe:
Made a PR for php 8 instead.

PR #1853, opened 2022-05-26
bind: add required dependency to DEPENDS

esselfe:
Build fails because of missing python-chardet.


esselfe:
Also missing: python-packaging

PR #1854, opened 2022-05-27
imlib2: version bumped to 1.9.0

esselfe:


lunar-linux-b01:
SUCCESS!

esselfe:
Seems like the imlib2-config script has been removed though nothing mentioned in the ChangeLog. Because of this giblib fails to configure.

Issue #1861, opened 2022-06-06
gst-libav:ninja: build stopped: subcommand failed.


lhw828:
Checking dependencies for gst-libav 
Building gst-libav version 1.20.2
The Meson build system
Version: 0.60.1
Source dir: /usr/src/gst-libav-1.20.2
Build dir: /usr/src/gst-libav-1.20.2/build
Build type: native build
Project name: gst-libav
Project version: 1.20.2
C compiler for the host machine: gcc (gcc 10.2.0 "gcc (GCC) 10.2.0")
C linker for the host machine: gcc ld.bfd 2.34
Host machine cpu family: x86_64
Host machine cpu: x86_64
Found pkg-config: /usr/bin/pkg-config (1.8.0)
Run-time dependency libavfilter found: YES 7.110.100
Run-time dependency libavformat found: YES 58.76.100
Run-time dependency libavcodec found: YES 58.134.100
Run-time dependency libavutil found: YES 56.70.100
Checking if "libav is provided by FFmpeg" with dependencies libavfilter, libavformat, libavcodec, libavutil: compiles: YES 
Has header "unistd.h" : YES 
Run-time dependency gstreamer-1.0 found: YES 1.20.0
Run-time dependency gstreamer-base-1.0 found: YES 1.20.0
Run-time dependency gstreamer-check-1.0 found: YES 1.20.0
Run-time dependency gstreamer-video-1.0 found: YES 1.20.0
Run-time dependency gstreamer-audio-1.0 found: YES 1.20.0
Run-time dependency gstreamer-pbutils-1.0 found: YES 1.20.0
Library m found: YES
Compiler for C supports arguments -fvisibility=hidden: YES 
Compiler for C supports link arguments -Wl,--exclude-libs=ALL: YES 
Compiler for C supports arguments -fno-strict-aliasing: YES 
Message: GStreamer debug system is disabled
Compiler for C supports arguments -Wmissing-declarations: YES 
Compiler for C supports arguments -Wmissing-prototypes: YES 
Compiler for C supports arguments -Wold-style-definition: YES 
Compiler for C supports arguments -Wredundant-decls: YES 
Compiler for C supports arguments -Wundef: YES 
Compiler for C supports arguments -Wwrite-strings: YES 
Compiler for C supports arguments -Wformat: YES 
Compiler for C supports arguments -Wformat-nonliteral: YES 
Compiler for C supports arguments -Wformat-security: YES 
[...]

PR #1862, opened 2022-06-06
gst-libav:depends elfutils

lhw828:
gst-libav:depends elfutils to fix https://github.com/lunar-linux/moonbase-other/issues/1861


lunar-linux-b01:
SUCCESS!

PR #1867, opened 2022-06-14
mailutils: version bumped to 3.15

esselfe:


esselfe:
So the CI fails because heirloom-mailx conflicts and is not uninstalled by default.
Not too sure what can be done about this.

PR #1888, opened 2022-08-10
fuse:bump to 3.11.0

lhw828:


lunar-linux-b01:
SUCCESS!

filesys/fuse/BUILD
```diff
@@ -1,10 +1,4 @@
-OPTS+=" --enable-lib \
-        --enable-util \
-        --disable-static \
-        --disable-example \
-        --sbindir=/usr/bin"
-
-default_build &&
+default_meson_build &&
 
 if module_installed systemd; then
```
Ratler, line 3: We can drop the if check for systems, since that would always be present.

That init file should always be removed :)

Ratler (changes requested):

lunar-linux-b01:
SUCCESS!

stumbles:
There are two fuse modules in other. Fuse and fuse-3 (currently 3.11.x). They are not the same. Last I tried a while back locally, using fuse at version 3.11.x caused some dependent modules not to work. Its been a while since testing this, so before bumping please thoroughly check dependent modules will still work.

PR #1889, opened 2022-08-11
gnupg:bump to 2.3.7

lhw828:


lunar-linux-b01:
SUCCESS!

Ratler:
The 2.3.x version is a development version, we should create a separate gnupg-dev module instead.

PR #1893, opened 2022-08-11
httpd: opt for the CDN instead

esselfe:
The main site mentions:  
So the CDN (Content Delivery Network)  link should be used instead.


lunar-linux-b01:
SUCCESS!

PR #1894, opened 2022-08-11
apr: use CDN link instead

esselfe:
The main site says:
So the CDN link should be used here too.


lunar-linux-b01:
SUCCESS!

PR #1895, opened 2022-08-11
apr-util: use CDN link instead

esselfe:
The main site says:
So use the CDN link instead.


lunar-linux-b01:
SUCCESS!

PR #1897, opened 2022-08-12
docbook-utils: fix broken URL

esselfe:
The URL from RedHat was broken, so use the URL that BLFS uses.


lunar-linux-b01:
SUCCESS!

PR #1904, opened 2022-08-16
glew:Add an alternate download address

lhw828:


lunar-linux-b01:
SUCCESS!

stumbles:
You have two problems here.

1. The github url needs to be SOURCE_URL_FULL=https://github.com/nigels-com/glew/archive/refs/tags/glew-2.2.0.tar.gz. However that's not needed since they provide a properlly named tarball, so nothing needed after the "tags". Even then that leads to problem 2.

2. There is a mismatch between the github tarball (.tar.gz) and the sourceforge tarball (tgz). While github is first it will always pull from there and never get to the sourceforge url unles github goes down or whatever. Then it will faill due to source name being different.

PR #1960, opened 2022-08-24
python-pytest-runner:6.0.0

lhw828:


esselfe:
Fails with this error message:
`python3: can't open file '/usr/src/pytest-runner-6.0.0/setup.py': [Errno 2] No such file or directory`

stumbles:
It requires the new method of installing python modules so our default_python_build function only works with those modules using setup.py.

stumbles:
Closing with commit; e04a8f0c5477a2dd6f99d86e9fe016484a242d5c

PR #1969, opened 2022-08-24
python-pyfits:bump to 3.5

lhw828:


esselfe:
Fails on my laptop.

stumbles:
Per the CHANGES.txt for that version;

This release officially deprecates PyFITS. Please use astropy.io.fits instead.

AFAICT no other mooonbase modules use this.

So I don't think its worth the time to fix it. Instead, create a python-astropy module using this https://www.astropy.org/ and will most likely require creating one or more additional python modules.

stumbles:
Closing with PR#3690 and this module will be removed.

PR #1991, opened 2022-08-25
python-scikit:bump to 1.1.2

lhw828:


esselfe:
Passed on my laptop.

Also 1.2.0 is out.

PR #2009, opened 2022-08-25
python2-ipy:bump to 1.0.1

lhw828:


esselfe:
Cannot fetch the source, the URL is broken.

PR #2035, opened 2022-08-28
mtools: fix typo in UPDATED

esselfe:
This was causing mtools to always be candidate in the update list when running `lunar update`


lunar-linux-b01:
SUCCESS!

PR #2041, opened 2022-09-03
php: bump to 8.1.10 and config option changes

esselfe:


esselfe:
About mcrypt deprecation: https://bugs.php.net/bug.php?id=73734

lunar-linux-b01:
SUCCESS!

PR #2094, opened 2022-09-16
virt-manager:bump to 4.1.0

lhw828:


esselfe:
Can't test because of circular dependencies in python modules.
I might retest if the python situation is fixed.

esselfe:
Ok I tested on my laptop and it works, I just don't get why take that URL, it should be
https://virt-manager.org/download/sources/virt-manager/virt-manager-4.1.0.tar.gz

stumbles:
Closing with commit 4de653ebd50912eab11133d51079219ca12071e4

PR #2097, opened 2022-09-16
libvirt-python:bump to 8.7.0

lhw828:


esselfe:
Fails at applying the patch, compiles fine without it.

esselfe:
Also 8.10.0 is out.

stumbles:
Closing with PR1850.

PR #2104, opened 2022-09-20
docker:bump to 20.10.18

lhw828:


esselfe:
The build fails here, I get:
Invalid commit for TINI_COMMIT, is 378bbbc8909a960e89de220b1a4e50781233a740 and should be 

Yes not the blank :P

Also 20.10.22 is out.

PR #2115, opened 2022-09-20
bochs:bump to 2.7

lhw828:


esselfe:
Fails with multiple errors.

PR #2127, opened 2022-09-20
r8168:bump to 8.050.03

lhw828:


esselfe:
On my laptop the last messages were:
make: *** /usr/src/linux-: No such file or directory.  Stop.
install: cannot stat 'src/*.ko.gz': No such file or directory

Probably because it's not adapted for my linux-lts module installed instead of linux...

esselfe:
Also 8.051.02 is out now.

PR #2147, opened 2022-09-21
synergy-core:bump to 1.14.5.23

lhw828:


stumbles:
Your attempting to rename the module to synergy-core and that doesn't exist. Leave the module name as is and readjust as needed.

esselfe:
Version 1.14.5.17 is out, I suggest a new PR, even a new module, so the module name is right.

stumbles:
Going on a limb here and ask the original author of this PR, how did you even compile this module before the PR creation? I have my suspensions and will leave it at that.

This version has switched to cmake making the BUILD useless.

PR #2155, opened 2022-09-21
xlayoutdisplay:bump to 1.4.0

lhw828:


esselfe:
This one fails because default_cmake_build is called but it's not a cmake-based project.
PRE_BUILD can be deleted and deleting BUILD also makes it go further, but still failing.
Now the errors are:
"Package libprocps was not found in the pkg-config search path."
"src/xutil.cpp:17:10: fatal error: proc/readproc.h: No such file or directory."

stumbles:
readproc.h is supposed to be provided by procps (core) however it does not get installed by the module. After testing procps to install its missing header xlayoutdisplay still fails because it does not know what to do with readproctab in src/xutil.cpp. 

So until the xlayoutdisplay dev fixes this to deal with "next gen" procps this module is done.

lhw828:
Retest this please.

stumbles:
Closing with;
https://github.com/lunar-linux/moonbase-other/commit/c483a49bd40329b5b459c1d9ad6c68dbd15ac4de

PR #2166, opened 2022-09-21
bashible:bump to 0.3.1

lhw828:


esselfe:
The build fails here, "cp: cannot stat '*.bash': No such file or directory"
So something is wrong in the BUILD.

stumbles:
Yes the BUILD is broken.

Delete the 

cp *.bash "/usr/share/${MODULE}/" &&

line.

You might want to add something like;

install -D -m755 ${MODULE}.* "/usr/bin/" &&

after the prepare_install so its .ble in TLD gets installed to /usr/bin. I'm guessing the TLD ble templates are needed to function properly. Never have used the module before so its a WAG.

lunar-linux-b01:
SUCCESS!

PR #2177, opened 2022-09-23
qpdf:fix the BUILD

lhw828:


lhw828:
install: cannot stat 'completions/bash/qpdf': No such file or directory

lunar-linux-b01:
SUCCESS!

PR #2190, opened 2022-10-21
fluxbox: fix pointer comparison to zero error

esselfe:
The build was giving:
`util/fluxbox-remote.cc:76:32: error: ordered comparison of pointer with integer zero ('unsigned char*' and 'int')
   76 |             && text_prop.value > 0`
so introduce a fix for comparison with NULL instead.


lunar-linux-b01:
SUCCESS!

PR #2194, opened 2022-10-26
ffmpeg:bump to 5.1.2

lhw828:


lunar-linux-b01:
SUCCESS!

stumbles:
How many other modules have you checked that this bump breaks? Even minor version bumps have and can break things and have seen others needing patches on such a major bump. Its the main reason I've not bumped this myself, waiting for others to catch up.

lhw828:
> How many other modules have you checked that this bump breaks? Even minor version bumps have and can break things and have seen others needing patches on such a major bump. Its the main reason I've not bumped this myself, waiting for others to catch up.

Just need someone has to take a step first.
At present, Lunar Linux can no longer be maintained in the regular development mode, and there are too few official maintenance personnel, just two people, or occasionally appear to make some updates and adjustments. There is a Chinese saying "不破不立"——
There's no making without breaking.

Of course, I still respect your professional decision. I'll just say a few thoughts.

PR #2199, opened 2022-10-29
Python-cryptography: moving it to moonbase-other

dagbrown:
It's only "needed" for one function used in one module in core (ca-certificates) and that module doesn't even do anything with what it uses python-cryptography for, so to make life simpler, I just removed the requirement for python-cryptography to be in core.

This is python-cryptography from core, and all of its dependencies.

### DIFF
--- a/python/python-asn1crypto/DEPENDS
+++ b/python/python-asn1crypto/DEPENDS
@@ -1,0 +1,1 @@
+depends python-setuptools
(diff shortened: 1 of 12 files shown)


lunar-linux-b01:
SUCCESS!

dagbrown:
reference to PR moving python-cryptography out of core: https://github.com/lunar-linux/moonbase-core/pull/3202

PR #2220, opened 2022-12-22
create-headers: Version bump to 6.0.11.

dagbrown:


stumbles:
I think you need to fix the source urls to https://cdn.kernel.org/pub/linux/kernel/v6.x/

PR #2226, opened 2023-01-13
python-pep517:just make a test

lhw828:
python-pep517:just make a test to confirm circular dependency


stumbles:
Closing with commit 61ce6364e8

PR #2235, opened 2023-02-02
apr-util: version bumped to 1.6.3

esselfe:


esselfe:
The CI failed because apr 1.7.0 cannot be retrieved, I'll wait for the CI to have 1.7.2.

PR #2241, opened 2023-02-05
postfix: Bump to 3.7.4

dagbrown:
Note that postfix-3.7.2 seems to have been deleted from postfix.org. I wonder what was wrong with it.


lunar-linux-b01:
SUCCESS!

PR #2242, opened 2023-02-05
mutt: change mime-types dependency to media-types

dagbrown:
I wonder if we can do some sort of automated monitoring for modules that depend on stuff that's been cratered


lunar-linux-b01:
SUCCESS!

PR #2247, opened 2023-02-19
php: version bumped to 8.2.3

esselfe:


esselfe:
lld fails configuring...

PR #2261, opened 2023-03-12
Linux LTS proposal

dagbrown:
There are now multiple different LTS branches of Linux--and also, the Lunar "stable" kernel is currently one of the LTS versions.

This takes the latest 6.1 release, which has been set to "long-term" by the kernel maintainers, and adds it as a new LTS kernel called "linux-6.1-lts". It also preserves the current LTS kernel, renaming it from "linux-lts" to "linux-5.15-lts".


esselfe:
I think it's a good idea :)

esselfe:
I've just thought about that... there's missing kernel-headers-6.1-lts and kernel-headers-5.15-lts

esselfe:
The CI fails because compressed firmware is not enabled in its kernel config.

PR #2264, opened 2023-03-27
rustc: Fix broken build when MAKES isn't set

dagbrown:
Default to 1 build at a time if MAKES isn't set to some other value, just like how the rest of the system works.


lunar-linux-b01:
SUCCESS!

PR #2268, opened 2023-04-03
qemu: Bumped to 7.2.0

dagbrown:
Changed a bunch of things because qemu's switched its build system to ninja. Fortunately the devs helpfully still included a "configure" script that writes a makefile so I didn't have to make quite such dramatic changes as I'd initially feared.


lunar-linux-b01:
SUCCESS!

PR #2272, opened 2023-04-16
New module: xorriso: an ISO9660/RockRidge filesystem manipulator

dagbrown:
How this wasn't in the moonbase already is a wonder to me. It's so very useful!


lunar-linux-b01:
SUCCESS!

PR #2282, opened 2023-05-14
lld: Another llvm test. In some PRs that pull in lld, cmake will tank.

stumbles:


stumbles:
it fails on GetErrcMessages, which is provided by /usr/lib/cmake/llvm/GetErrcMessages.cmake (llvm). The llvm CI compile log show that file being "installed".

So what gives?

stumbles:
Retest this please.

PR #2284, opened 2023-05-15
llvm: Tweaking the BUILD. Don't think it will correct lld tanking bec…

stumbles:
…ause

cmake due to camke not finding an installed llvm file, but CI needs something to do.


lunar-linux-b01:
SUCCESS!

PR #2286, opened 2023-05-17
llvm: Let's try skipping rpaths and

stumbles:
let's put default_python3_build back in there (OPTS needs unset) else the -D of cmake confuses python.


lunar-linux-b01:
SUCCESS!

PR #2293, opened 2023-06-06
heimdal: Bump to 7.8.0. Many changes here. The relevant comments are …

stumbles:
…in the BUILD

The plan is remove heimdal conflicts from crypto/krb5/CONFLICTS, remove heimdal from other/aliases.


lunar-linux-b01:
SUCCESS!

Ratler (changes requested):
I don't agree with moving things to a heimdal prefix, this will break a lot of things for existing scripts. If we need to move something, we should move krb5 to its own prefix.

This change will also completely break existing KDCs, since the database location was changed. I believe these changes should be reverted, and only keep the bump.

lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

PR #2301, opened 2023-06-24
camlp4: Bump to 5.0.0+1. Yes its still in 4.x but adjusted to compile

stumbles:
with the current ocaml 5.x.


lunar-linux-b01:
SUCCESS!

PR #2322, opened 2023-06-29
ImageMagick:bump to 7.1.1-12

lhw828:
ImageMagick:bump to 7.1.1-12 and fix the download link,otherwise the original link will not be available when the new version appears.


lunar-linux-b01:
SUCCESS!

PR #2500, opened 2023-08-01
SDL_sound:bump to 2.0.2

lhw828:


esselfe:
This module changed to the cmake build system.

lunar-linux-b01:
SUCCESS!

esselfe:
Looks like this module should be renamed SDL2_sound now...

PR #2507, opened 2023-08-01
vmpk:bump to 0.8.9

lhw828:


esselfe:
Can't test because qt5 5.15.10 fails here.

esselfe:
This version wants drumstick 2.8 but it fails to build.

stumbles:
Drumstick is bumped. But fails on meson when trying to compile pulseaudio. Merging anyway.

PR #2514, opened 2023-08-02
OpenImageIO-oiio:bump to 2.4.14.0

lhw828:


esselfe:
Can't test because qt5 5.15.10 fails to build here.

esselfe:
Ok so qt5 defaults to y, but when selecting no I get this:

CMake Error at CMakeLists.txt:84 (message):
  Not allowed to run in-source build!


-- Configuring incomplete, errors occurred!

PR #2522, opened 2023-08-02
geeqie:bump to 2.1

lhw828:


esselfe:
The patch is broken.

esselfe:
Now it fails with multiple linker "undefined reference" for lua_* symbols.

stumbles:
Closing. See PR#3347.

PR #2523, opened 2023-08-02
exiv2:bump to 0.28.0

lhw828:


esselfe:
Not merging, breaks gexiv2

PR #2525, opened 2023-08-02
cegui:bump to 1.1.2

lhw828:


esselfe:
The URL is broken.

esselfe:
After correcting the hash turns out it extracts to /usr/src/ceed-cpp-1.1.2

esselfe:
Well there's no configure script, no cmake files, no meson either... someone gotta find how to build this :P

stumbles:
That's not the right source url link for cegui.

stumbles:
The correct url for cegui download is; https://github.com/cegui/cegui/ and their current version is the same as that in moonbase, 0.8.7.

Closing; invalid.

PR #2532, opened 2023-08-02
freedroidRPG:bump to 1.0

lhw828:


esselfe:
Fails with:

system.h:115:10: fatal error: SDL.h: No such file or directory

lunar-linux-b01:
SUCCESS!

PR #2533, opened 2023-08-02
freedink:bump to 109.6

lhw828:


esselfe:
This project changed from SDL to SDL2. Fails with:

Errors and warnings recap:
* You need to install the 'cxxtest' testing framework:
Could not find 'cxxtestgen'

See https://github.com/CxxTest/cxxtest

esselfe:
With this fix commit it builds fine, but when you start it with 'freedink -r /usr/share/games/freedink' it crashes the Xorg server! Not to be merged at this point then.

esselfe:
The patch is taken from https://mail.gnu.org/archive/html/bug-freedink/2022-07/msg00001.html

PR #2572, opened 2023-08-03
openafs:bump to 1.8.10

lhw828:


esselfe:
Can't test because openafs-driver is required and it's broken.

Ratler:
This PR would break openafs completely on existing installations. Just commenting out various patches without understanding why they are there should be avoided. I will close this PR and bump openafs myself.

PR #2573, opened 2023-08-03
openafs-driver:bump to 1.8.10

lhw828:


esselfe:
Even after correcting PRE_BUILD I get:
"checking how to set cflags for linux kbuild... configure: error: cannot find way to set linux kbuild cflags"

stumbles:
Closing with PR#4764

PR #2586, opened 2023-08-03
xemacs:bump to 21.5.35

lhw828:


esselfe:
The patch is broken.

stumbles:
That's a beta version. Do you really want that? We try to avoid using beta versions of things unless there is a very compelling reason.

PR #2604, opened 2023-08-04
semantik:bump to 2.0.25

lhw828:


stumbles:
Where are you getting that version? Their home page and gitlab only shows 1.2.10 as the latest.

PR #2609, opened 2023-08-04
lilypond:bump to 2.25.7

lhw828:


stumbles:
That's an unstable version, and we try to avoid such unless there is an absolute need vis-à-vis, security, failure to build, etc.

Closing.

PR #2627, opened 2023-08-08
bashdb:bump to 5.0-1.1.2

lhw828:


esselfe:
Fails with "configure: error: This package is only known to work with Bash 5.0"

stumbles:
Closing with PR#3324.

PR #2657, opened 2023-08-08
lttng-modules:bump to 2.13.10

lhw828:


stumbles:
Compiles fine here with 6.1.60-lts.

PR #2735, opened 2023-08-10
newsboat:bump to 2.32

lhw828:


lhw828:
Needed stfl and wait https://github.com/lunar-linux/moonbase-other/pull/2767

PR #2743, opened 2023-08-10
bind:bump to 9.19.15

lhw828:


lhw828:
Needed "python-idna",but python-idna depends python-flit-core
https://github.com/lunar-linux/moonbase-other/pull/2768

PR #2778, opened 2023-08-11
spamassassin:bump to 4.0.0

lhw828:


lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

esselfe:
Fails with the tests enabled but succeeds if they're disabled.

PR #2782, opened 2023-08-11
mutt:bump to 2.2.10

lhw828:


lhw828:
depends media-types but media-types 9.0.0 is not exist
wait for https://github.com/lunar-linux/moonbase-other/pull/2784

PR #2785, opened 2023-08-11
mailutils:bump to 3.16

lhw828:


lhw828:
mailutils: Install & use optional (not installed) module guile
          Purpose: for Guile interface (unmaintained, so not recommended) ? [n] Checking dependencies for mailutils 
]0;a4dff065fdf8 lin[24362]: compiling mailutils WARNING: heirloom-mailx conflicts with module mailutils
mailutils: Do you want to remove module heirloom-mailx ?  [n] ERROR: Cannot continue installing conflicting modules, exiting!
Build step 'Execute managed script' marked build as failure
Discard old builds...
#3198 is removed because old than numToKeep

PR #2804, opened 2023-08-14
sysvinit:bump to 3.04

lhw828:


esselfe:
Fails with missing $SCRIPT_DIRECTORY/init.d

esselfe:

PR #2807, opened 2023-08-14
openrc:bump to 0.48

lhw828:


esselfe:
Fails with "sed: Can't read mk/sys.mk: No such file or directory"

esselfe:

PR #2810, opened 2023-08-14
elogind:bump to 252.9

lhw828:


esselfe:
Can't test because eudev is broken.

esselfe:

PR #2825, opened 2023-08-14
virtuoso:bump to 7.2.10

lhw828:


esselfe:
Though perl and python support is broken...

PR #2848, opened 2023-08-14
rofi-calc：bump to 2.2.0

lhw828:


lhw828:
checking for dlopen... no
checking for dlopen in -ldl... yes
checking for cargo... cargo
checking for rustc... rustc
configure: error: librsvg needs at least rustc version 1.64.0

Creating /var/log/lunar/compile/librsvg-2.56.0.bz2

PR #2972, opened 2023-08-16
lfs:bump to 2.7.2

lhw828:


lhw828:
2023-08-16 18:22:11 (10.2 MB/s) - '/tmp/lfs-2.7.2.tar.gz' saved [3495940]

Building lfs version 2.7.2
error: could not find `Cargo.toml` in `/usr/src/lfs-2.7.2` or any parent directory
Creating /var/log/lunar/compile/lfs-2.7.2.bz2 
! Problem detected during BUILD

esselfe:
I think this module is gradually beging renamed to dysk.

PR #2995, opened 2023-08-16
rcm:bump to 1.3.6

lhw828:


esselfe:
Even after adding ./autogen.sh in BUILD this module wants the pystache python module which doesn't exist in the moonbase yet.

esselfe:
Also the hash is wrong.

stumbles:
Closed by PR #3136.

PR #3008, opened 2023-08-16
mpv:bump to 0.36.0

lhw828:


esselfe:
The CI failed with:

Unable to find development files for libass, or the version found is too old. Aborting.

PR #3032, opened 2023-08-25
openttd: move unpack ops into BUILD

esselfe:
Strangely the unpacking of the game files seem to be ignored if put in INSTALL, so move everything in BUILD.
Note that an attempt was made to put the unpack calls in POST_INSTALL, and it works, but installwatch doesn't track those at this point and it was breaking resurrection of the module. Now it works :)

### DIFF
--- a/games/openttd/BUILD
+++ b/games/openttd/BUILD
@@ -1,1 +1,13 @@
-default_cmake_build
+default_cmake_build &&
+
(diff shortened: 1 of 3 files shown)


lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

PR #3033, opened 2023-08-26
go: fix missing go.env

esselfe:
Stuff like containerd fail to build without it.


stumbles:
I thought dagbrown had this fixed.

PR #3062, opened 2023-09-03
gdbus-codegen: fix BUILD

esselfe:


esselfe:
This is related to the new meson build dir change in https://github.com/lunar-linux/lunar/commit/9b23f9fa61118f18f65e01b5976946192f7f3df7

PR #3076, opened 2023-09-25
gdbus-codegen: Bump to 2.78.0.

stumbles:


stumbles:
Don't waste time on this module. The latest version of glib-2 dropped the pythonism in gio/gdbus-2.0/codegen to generate gdbus-codegen in favor of meson. The meson.build located in that directory isn't meant to be run directly, that is, you need to do it from TLD. Unless I am blind, there are no meson.options to compile just gdbus.

The plan is to drop this module, change all modules dependent on gdbus to glib-2. In order for that to happen, the wait is on the glib-2 PR in core.

stumbles:
This module has been deleted in favor of glib-2. See #3390 in core.

PR #3080, opened 2023-09-27
fontforge: add patch for new gettext 0.22

esselfe:


esselfe:
See https://github.com/fontforge/fontforge/issues/5251

esselfe:
The CI is broken: cp: cannot stat '/zlocal/*': No such file or directory

PR #3082, opened 2023-09-28
birdtray: fix wrong source naming

esselfe:


esselfe:
The CI failed at pulseaudio, I'll merge since it's just a tarball renaming and it succeeded here.

PR #3086, opened 2023-09-30
geoclue: Bump to 2.7.1, adding some missing depends.

stumbles:


stumbles:
Fails casue CI doesn't compile meson for depends libsoup-3. Merging anyway.

PR #3092, opened 2023-10-01
Lhw828 master 464

stumbles:


stumbles:
Fails on pulseaudio.

esselfe:
Here it fails with "Missing configure and/or Makefile!"

esselfe:
Needs a "TYPE=cmake" in DETAILS, but then it fails with:

CMake Error at src/sdl-test-ui/CMakeLists.txt:19 (target_link_libraries):
  Target "projectM-Test-UI" links to:

    SDL2::SDL2main

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

stumbles:
Closing with https://github.com/lunar-linux/moonbase-other/commit/c02fa7040f6354c15605f7800c5c1b3806f98364

PR #3093, opened 2023-10-01
OpenBLAS: Adding. Will be a depends for armadilo bump.

stumbles:
Armadilo looks for atlas and/or OpenBLAS. In its compile log it is clear atlas will be depreciated and if both is present will ignore atlas in favor of OpenBLAS. No switches are needed.


lunar-linux-b01:
SUCCESS!

PR #3094, opened 2023-10-01
armadillo: Bump to 12.6.4 and adding OpenBLAS depends.

stumbles:


stumbles:
Fails because CI gcc is compiled without g77.

PR #3095, opened 2023-10-02
wlroots: Adding a patch and removing -D xwayland=disabled switch.

stumbles:


stumbles:
Fails on wayland CI being to old even though moonbase version is 1.22.0.

PR #3098, opened 2023-10-02
pocketsphinx: Bump to 5.0.2 and convert to cmake.

stumbles:


stumbles:
Fails compiling gstreamer; meson.build:1:0: ERROR: Meson version is 0.60.1 but project requires >= 0.62

PR #3100, opened 2023-10-06
gnupg: Bump to 2.4.3, Removed and add some openldap patches, and one …

stumbles:
…for emacs.

Deleting PRE_BUILD.


stumbles:
Even though libgpg-error is in DEPENDS, CI doesn't want to compile it. Merging this anyway.

PR #3106, opened 2023-10-10
cargo-c: Bump to 0.9.26.

stumbles:


stumbles:
I'll assum this; "
  The package requires the Cargo feature called `strip`, but that feature is not stabilized in this version of Cargo (1.58.0 "

refers to CI rustc  being to old.

PR #3113, opened 2023-10-14
sway: version bumped to 1.8.1

esselfe:


esselfe:
Installing with SUID bit set give an error message. The proper way to start sway is:
`dbus-run-session sway`

esselfe:
@stumbles can you confirm this is working with the current wlroots?

stumbles:
compiles fine here with current wlroots. Merging.

PR #3134, opened 2023-10-16
egl-wayland: Adjusting BUILD.,

stumbles:


stumbles:
The fail actually starts with spirv-tools but for some reason CI just keeps going. Bumped spriv-tools and lets see what happens now.

stumbles:
Retest this please.

PR #3135, opened 2023-10-16
libprojectM: Bump to 4.0.0. Now uses cmake.

stumbles:
Adjusted PRE_BUILD to see if it really uses/needs qt5. Greping the compile for qt revealed no results. We shall see.


lunar-linux-b01:
SUCCESS!

PR #3141, opened 2023-10-17
synergy: Bump to 1.14.6.19.

stumbles:


stumbles:
Ah, another to fail on pulseaudio.

PR #3156, opened 2023-10-21
libreoffice-bin: version bumped to 7.6.2

esselfe:


esselfe:
The CI failed at poppler.

PR #3157, opened 2023-10-21
poppler: Add missing depends nss as called out by CI.

stumbles:


stumbles:
Now fails at gnupg not finding  libgpg-error  even though  libgpg-error  is in DEPENDS as a depends.

PR #3158, opened 2023-10-22
sphinxtrain: Bump to 5.0.0. Converted to cmake.

stumbles:
Depends sphinxbase is depreciated and will be moved to crater. This module now depends on pocketsphinx.


stumbles:
Tanks on gstreamer; 

meson.build:1:0: ERROR: Meson version is 0.60.1 but project requires >= 0.62

PR #3161, opened 2023-10-23
bind-utils: install missing libs

esselfe:
The utilities fail to execute because of missing shared libraries. Taken from the LinuxFromScratch book (BLFS-dev).


lunar-linux-b01:
SUCCESS!

PR #3170, opened 2023-10-23
cbindgen: Bump to 0.26.0.

stumbles:


stumbles:
error: package `cbindgen v0.26.0 (/usr/src/cbindgen-0.26.0)` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.58.1-dev

PR #3172, opened 2023-10-23
wmenu: new module, an application menu for Sway

esselfe:


lunar-linux-b01:
SUCCESS!

esselfe:
The README says docs are optional but I haven't found a way to turn it off...

PR #3177, opened 2023-10-24
clamav: version bumped to 0.103.10

esselfe:


esselfe:
I might upgrade this module to the latest lts, 1.0.3, which uses cmake instead of autotools.

lunar-linux-b01:
SUCCESS!

PR #3180, opened 2023-10-24
gstreamer: Disabling docs and gobject-instrospetion may as well

stumbles:
be a depends at this point.


stumbles:
meson.build:1:0: ERROR: Meson version is 0.60.1 but project requires >= 0.62

PR #3182, opened 2023-10-25
foomatic-db-engine: Bump to 4.0-20231024. We don't use any of the

stumbles:
"current" foomatics because they change often. This would cause un-needed effort maintaining them, unless you disable sha256, which we prefer *not* to do.

If you want them bumped, a copy needs to be uploaded to mirrors on robusta. For those without acces, I'm generally around so message if in need when you decide to bump any of the foomatic stuff.


stumbles:
Another failure because CI won't compile/install libgpg-error  for gnupg even though it's in DEPENDS for gnupg.

PR #3190, opened 2023-10-26
foomatic-db-nonfree: Bump to 20231025.

stumbles:


stumbles:
Again, CI refuses to compile/install libgpg-error even though its in the gnupg DEPENDS. 

Merging anyway.

PR #3199, opened 2023-10-26
linux-lts: version bumped to 6.1.60

esselfe:


esselfe:
The CI failed at linux-firmware with:

The kernel option CONFIG_FW_LOADER_COMPRESS is not enabled.
Please reconfigure and rebuild your kernel. Then try installing linux-firmware again.

PR #3237, opened 2023-10-28
conky: version bumped to 1.19.6 + Wayland support

esselfe:


esselfe:
I tried with Wayland enabled in sway but it still gives me the windowed version, I haven't found a way to render straight on the desktop like it does with Xorg...

lunar-linux-b01:
SUCCESS!

PR #3241, opened 2023-10-29
SuiteSpare: Bump to 7.3.0.

stumbles:


stumbles:
This and other modules requiring gfortran/g77 and its ilk will never compile successfully, since the base gcc on CI and moonbase does not have it enabled.

PR #3266, opened 2023-10-31
lzma-sdk: Bump to 2301 and some BUILD adjustments.

stumbles:


stumbles:
/var/lib/lunar/moonbase/zlocal/lzma-sdk/PRE_BUILD: line 4: /usr/bin/7za: No such file or directory

If only CI would actually do what's in DEPENDS, ie in this case, compile p7zip.

lunar-linux-b01:
SUCCESS!

PR #3269, opened 2023-10-31
darktable: Bump to 4.4.2. Adding two patches, notable one from the

stumbles:
gentoo folks dealing with json stuff and one from arch folks dealing with with libavif.

### DIFF
--- a/graphics/darktable/DEPENDS
+++ b/graphics/darktable/DEPENDS
@@ -1,4 +1,5 @@
 # hard requirements
+depends python-setuptools
 depends sqlite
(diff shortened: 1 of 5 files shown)


stumbles:
Right then. The core PR adding python-setuptools to python-markupsafe was merged, yet CI claims no setuptools module found.

stumbles:
Now fails on; configure: error: librsvg needs at least rustc version 1.70.0

CI rustc to old.

esselfe:
Can't test because python-referencing seems to have disappeared from the moonbase...

PR #3301, opened 2023-11-03
python-scipy: Bump to 1.11.3.

stumbles:


stumbles:
Fails on lapack; g77 is required to install this module. Rebuild gcc with it enabled.

PR #3302, opened 2023-11-03
python-pythran: Bump to 0.12.0. Also testing DEPENDS to see if

stumbles:
those depends are really needed.


lunar-linux-b01:
SUCCESS!

PR #3307, opened 2023-11-06
git: semi-cosmetic fix to DEPENDS file

dagbrown:
Since the `depends.cache` file uses colons as field separators, colons in the optional_depends explanations confuses anything trying to parse it.


lunar-linux-b01:
SUCCESS!

PR #3313, opened 2023-11-21
git: version bumped to 2.43.0

esselfe:
Weirdly the build process fails if passing --with-libpcre2 instead of --with-libpcre


lunar-linux-b01:
SUCCESS!

PR #3317, opened 2023-11-23
claws-mail: version bumped to 4.2.0 + new optional deps

esselfe:


esselfe:
The CI failed at gnupg because libgpg-error is too old on it.

PR #3346, opened 2023-12-13
owncloudclient: Bump to 5.2.0.12726. It now depends on qt6. Making

stumbles:
adjustments to BUILD and DEPENDS.


stumbles:
libsecret fails with the meson error;

importlib.metadata.PackageNotFoundError: No package metadata was found for meson

PR #3347, opened 2023-12-16
geeqie: Test the lua patch.

stumbles:


stumbles:
As expected. CI is out of date and fails on gnupg needing libgpg-error > 1.46. Current moonbase version is 1.47.

PR #3349, opened 2023-12-16
gimp: version bumped to 2.10.36

esselfe:


esselfe:
The CI failed because its meson is too old for json-glib

PR #3350, opened 2023-12-16
php: version bumped to 8.3.0

esselfe:


esselfe:
The CI failed at cargo-c because cargo is too old...

PR #3354, opened 2023-12-21
php: version bumped to 8.3.1

esselfe:


esselfe:
The CI failed because cargo is too old.

PR #3355, opened 2023-12-23
postfix: Version bumped to 3.8.4

dagbrown:
This includes a fix for https://www.postfix.org/smtp-smuggling.html wherein an attacker can fool your mail server into delivering emails to where it's not supposed to, via the magic of poorly-formed line endings.


lunar-linux-b01:
SUCCESS!

PR #3374, opened 2024-01-21
xsel: fix hash and prevent double configure

esselfe:
The configure script was running twice, so fix that with autoreconf.


lunar-linux-b01:
SUCCESS!

PR #3383, opened 2024-01-23
MPlayer: Bump to 1.5. Removed MPlayer-all and MPlayer-essentials. Per

stumbles:
the changelog live555 is disabled by default and scheduled to be removed. Also in the BUILD, mplayer has always been a bit touchy about compile options and unsetting compiler wasn't enough, just remove $OPTS.

Added a couple of patches from the LFS folks. Thank you guys.

This will no doubt fail on CI.


lunar-linux-b01:
SUCCESS!

PR #3397, opened 2024-02-01
egl-wayland: Adding a depends.

stumbles:


stumbles:
Fails as before. Wonder if this is a kernel version thing CI is using.

PR #3409, opened 2024-02-03
neomutt: version bumped to 20240201

esselfe:


esselfe:
The CI failed at gnupg because libgpg-error is too old.

PR #3441, opened 2024-02-10
pigz: version bumped to 2.8

esselfe:
pushd and popd add unnecessary output, so reduce noise with a simple ln command.


lunar-linux-b01:
SUCCESS!

PR #3442, opened 2024-02-10
rdiff-backup: version bumped to 2.2.6

esselfe:


esselfe:
The CI failed the first time because pip was missing, then when installing pip it says:

FileExistsError: File already exists: /usr/bin/pip3.9

PR #3444, opened 2024-02-13
lapack: Bump to 3.12.0.

stumbles:


stumbles:
CI has no gcc with g77, thus fail.

PR #3586, opened 2024-02-19
python-scipy:bump to 1.12.0

lhw828:


lhw828:
Building lapack version 3.12.0
g77 is required to install this module. Rebuild gcc with it enabled.

Creating /var/log/lunar/compile/lapack-3.12.0.bz2 
! Problem detected during PRE_BUILD

PR #3587, opened 2024-02-19
python-scikit:bump to 1.4.1

lhw828:


lhw828:
2024-02-19 05:03:29 (21.7 MB/s) - '/tmp/lapack-3.12.0.tar.gz' saved [7933607/7933607]

Building lapack version 3.12.0
g77 is required to install this module. Rebuild gcc with it enabled.

Creating /var/log/lunar/compile/lapack-3.12.0.bz2 
! Problem detected during PRE_BUILD

PR #3610, opened 2024-02-19
dav1d: Bump to 1.4.0 and adding a POST_INSTALL to

stumbles:
lunar fix modules that use it if installed.


lunar-linux-b01:
SUCCESS!

PR #3635, opened 2024-02-24
Swap iw and wireless_tools

dagbrown:
iw is actively maintained, and the last release of wireless_tools was in 2010. So iw shouldn't be in other any more, and wireless_tools should take its place.


lunar-linux-b01:
SUCCESS!

PR #3684, opened 2024-03-02
freshplayer: Bump to 0.3.11. Reworked DEPENDS and adjusted BUILD.

stumbles:
Per; https://github.com/i-rinat/freshplayerplugin/issues/389, disabling hardware decoding.

If and when this module breaks again I highly recommend it be sent off to /dev/null, it doesn't deserve crater.


lunar-linux-b01:
SUCCESS!

PR #3768, opened 2024-03-28
zziplib: fix wrong libs in pkgconfig file

esselfe:
The .pc files seem to be incorrectly generated. The installed libs end in -0.so so add a temporary fix until this situation is resolved upstream.


lunar-linux-b01:
SUCCESS!

PR #3770, opened 2024-03-31
other/crater: add a basic README

esselfe:
The creation of this file is mostly justified to keep the directory present even if there's no modules in it. And it doesn't hurt to add a little description to it :)


lunar-linux-b01:
SUCCESS!

PR #3828, opened 2024-04-29
kexec-tools: version bumped to 2.0.28 + new patch

esselfe:
The build was failing with:

purgatory/arch/i386/entry32-16.S:23: Error: 64bit mode not supported on `i386'.


lunar-linux-b01:
SUCCESS!

PR #3830, opened 2024-05-06
zfs: Version bumped to 2.2.4

dagbrown:
This includes support for Linux 6.8 kernels, so it'll be nice for it to work with a recent daily ISO


lunar-linux-b01:
SUCCESS!

PR #3862, opened 2024-07-07
linux-lts: version bumped to 6.6.37

esselfe:


stumbles:
Fails on; The kernel option CONFIG_FW_LOADER_COMPRESS is not enabled.

I presume it refers to CI kernel.

esselfe:
Closing with https://github.com/lunar-linux/moonbase-other/pull/3872

PR #3868, opened 2024-07-12
create-headers: version bump + add rsync as dep

esselfe:


esselfe:
Well it failed with:

Warning! This module does not need to be run on every kernel update. The optimum time would be from one major kernel version to the next.
 
create-headers: Do you wish to continue (y/n)?  [n]

PR #3872, opened 2024-07-13
linux-lts: version bumped to 6.6.39

esselfe:


esselfe:
Failed with:

The kernel option CONFIG_FW_LOADER_COMPRESS is not enabled.

esselfe:
I'll merge anyways since it's just a version bump, and it fails because of a CI misconfiguration.

PR #3876, opened 2024-07-13
rustc: Adding a config.toml. Started with libopenraw failing on;

stumbles:
error[E0635]: unknown feature `stdsimd`

The sedit that was in BUILD about download-ci-llvm is now in config.toml. Additionally seems we were not actually using stable. The locked-deps and vendor entries puts us on stable.

Anyway, fixes libopenraw.


lunar-linux-b01:
SUCCESS!

PR #3971, opened 2024-07-26
mongodb: Bump to 7.0.12. Removed boost from depends, current version …

stumbles:
…removes

boost/filesystem/convenience.hpp and needed by mongo. Till then we will use their version.


lunar-linux-b01:
SUCCESS!

PR #4029, opened 2024-07-31
xinput_calibrator: version bumped to 0.8.0

esselfe:
This is from the email I got:

```
This is the first release of xinput_calibrator after the move from
https://github.com/tias/xinput_calibrator to our gitlab instance. The
aim of this move is so this little tool, still in use here and there,
can see some of the benefits of group maintainership and doesn't go
completely stale. Don't expect active maintainership or new features
being added, we're just trying to keep this on life support.

The git shortlog below is for the ~14 years of patches since the 0.7.5
release, the last 10 of which the project was sitting idle.
```

Also the new URL is taken from this email.

### DIFF
--- a/x11-utils/xinput_calibrator/BUILD
+++ b/x11-utils/xinput_calibrator/BUILD
@@ -1,4 +1,5 @@
 OPTS+=" --with-gui=x11"
 
-./autogen.sh &&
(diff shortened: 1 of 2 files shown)


lunar-linux-b01:
SUCCESS!

PR #4039, opened 2024-07-31
python-pkgconfig: Adding, will be a depends for python-faust-cchardet

stumbles:
which will be a depends for the coming calibre update.


lunar-linux-b01:
SUCCESS!

PR #4058, opened 2024-07-31
antiword: Chagning SOURCE and WEB_SITE urls since the orignal seems t…

stumbles:
…o have

gone away. The sources were chosen based on listed by ugrep.


lunar-linux-b01:
SUCCESS!

PR #4067, opened 2024-08-01
xca: Bump to 2.6.0. Adjusted BUILD should Sphinx be already installed…

stumbles:
… avoid

the need for all those sphinxcontrib modules.


lunar-linux-b01:
SUCCESS!

PR #4078, opened 2024-08-01
googletest: Bump to 1.15.2. Gmock is now part of googletest, so the m…

stumbles:
…odule

gmock will get removed, eventually.


lunar-linux-b01:
SUCCESS!

PR #4097, opened 2024-08-04
wireless-regdb: version bumped to 2024.07.04 + fix wrong path

esselfe:
This module was installing in /usr/usr/lib/firmware :P


lunar-linux-b01:
SUCCESS!

esselfe:
I also tried putting regulatory.db in /lib/firmware but I'm still getting error -2 (No such file or directory) in the systemd journal. You can check this with 'journalctl -b | grep -C8 regulatory.db'.

PR #4110, opened 2024-08-05
CPU-X: version bumped to 5.0.4

esselfe:
The new URL is a Github redirect from the old one. The new website URL is taken from the new Github profile description.


lunar-linux-b01:
SUCCESS!

PR #4169, opened 2024-08-17
python-pypng: Adding, will be a depends for another new python

stumbles:
module called python-qrcode which will be a depends for a future bump of gajim.


lunar-linux-b01:
SUCCESS!

PR #4199, opened 2024-08-19
Move broken init systems to crater

esselfe:
Move elogind, eudev, openrc, openrc-elogind, openrc-eudev and sysvinit to crater.
The thing is when installing those it removes systemd and it's required for e2fsprogs, btrfs-progs and lvm2 now. If we really want to keep it then we need new modules, e.g. e2fsprogs-sysv with an older version supporting optional systemd support.

Another fact is that changing init system is kinda crazy to tell honestly, having tried it before, it's a little rock n' roll. This kind of project should be left entirely in the hands of expert users.


esselfe:
Closing with 1e741871190825d8927be9310394ee4332a4cffe d89d79824c7a7a4abdffd947aeaafdaee8b53be8 759ae35b2f944501d6ccd308959a664be916bdd0 04d67bcd68e7e994b5d3db02032c5085e55d5599 8cb9ffeeb66388e78ed075fdf9a02994b33e964d and 5dc6a6a3c83ef599b2269ac8767602c51a5543cc

PR #4226, opened 2024-08-21
colord: A few changes to DEPENDS and PRE_BUILD. Primarily in

stumbles:
PRE_BUILD adjusting its service file.


lunar-linux-b01:
SUCCESS!

PR #4246, opened 2024-08-23
gh: version bumped to 2.55.0

esselfe:
With the previous PRE_BUILD setup there was a leftover directory in /usr/src once lin is finished (cli-2.55.0).


lunar-linux-b01:
SUCCESS!

PR #4277, opened 2024-08-25
bochs: Bump to 2.8.

stumbles:


stumbles:
This module will never pass because of;

`configure: error: PCI host device mapping requires Linux as host`

PR #4317, opened 2024-08-30
libnvme: version bumped to 1.10

esselfe:
Next will come nvme-cli, it just needs the CI to have this merged first otherwise it will fail with libnvme 1.8


lunar-linux-b01:
SUCCESS!

PR #4319, opened 2024-08-30
php: version bumped to 8.3.11

esselfe:
Also change tabs for spaces.

PR #4320, opened 2024-08-30
xmlto: fix build with gcc 14 (Gentoo patch)

esselfe:
As per the CI failure seen from https://github.com/lunar-linux/moonbase-other/pull/4319


lunar-linux-b01:
SUCCESS!

PR #4324, opened 2024-08-30
Vulkan-Headers: Bump to 1.3.295 and adding a BUILD to use Ninja

stumbles:
as the cmake generator type.


lunar-linux-b01:
SUCCESS!

PR #4335, opened 2024-09-04
opensp: add Gentoo patch for newer clang

esselfe:
The error I was getting in configure was:

checking size of size_t... configure: error: cannot compute sizeof (size_t), 77


lunar-linux-b01:
SUCCESS!

PR #4348, opened 2024-09-11
pygtk: fix build with gcc 14

esselfe:
The error I was having was:

pango.c:5382:16: error: returning 'int' from a function with return type 'PangoFontMetrics *' {aka 'struct _PangoFontMetrics *'} makes pointer from integer without a cast [-Wint-conversion]
 5382 |         return pango_font_metrics_new();

This fix was found in the messages of the Arch pygtk package.

### DIFF
--- a/python/pygtk/BUILD
+++ b/python/pygtk/BUILD
@@ -2,7 +2,7 @@ export PYTHONDONTWRITEBYTECODE=1 &&
 
 OPTS+=" --prefix=/usr --enable-thread --disable-docs" &&
 
-./configure $OPTS PYTHON=python2 &&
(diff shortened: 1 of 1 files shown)


esselfe:
See comments at https://aur.archlinux.org/packages/pygtk

esselfe:
The CI failed at XML-SAX-Expat with:

Can't locate XML/Parser.pm in INC (you may need to install the XML::Parser module)

Note that XML-Parser is already in the DEPENDS.

stumbles:
Retest this please.

stumbles:
Retest this please.

PR #4354, opened 2024-09-16
efitools: Add another patch file to deal with gcc 14

dagbrown:
gcc 14 is a lot stricter than gcc 13, so we have to deal with it until upstream gets around to it.

(Surely there's a better way to deal with the strptime issue than what I did though...)


lunar-linux-b01:
SUCCESS!

PR #4382, opened 2024-09-23
ripgrep: version bumped to 14.1.1

esselfe:
Weird thing is 'rg --generate=...' shell completion generation doesn't seem to work in BUILD but does on the CLI.


lunar-linux-b01:
SUCCESS!

PR #4425, opened 2024-10-02
linux-lts: version bumped to 6.6.53

esselfe:


esselfe:
The CI failed because compressed firmware is not enabled in the kernel config.

esselfe:
Closing with https://github.com/lunar-linux/moonbase-other/pull/4482

PR #4426, opened 2024-10-02
neomutt: Bump to 20241002.

stumbles:


stumbles:
Fails do to w3m failing, which fails because CI doesn't want to install a specified depends for w3m, thus failure.

esselfe:
Retest this please.

PR #4427, opened 2024-10-02
w3m: Add missing depends.

stumbles:


stumbles:
Just another situation CI doesn't want to install a depends for whatever reason even when the depends is specified in the modules DEPENDS.

PR #4433, opened 2024-10-03
python-scipy: Bump to 1.14.1.

stumbles:


stumbles:
By default gcc does not enable g77, so is it that much an issue to have it a default?

For now I will merge this.

PR #4436, opened 2024-10-04
wireguard-tools: Extra tools to manage WireGuard VPNs

dagbrown:
It's worth noting that wireguard itself is now a standard part of the kernel, so there's no need for a separate module for that.


lunar-linux-b01:
SUCCESS!

PR #4453, opened 2024-10-11
bleachbit: version bumped to 4.6.2

esselfe:


esselfe:
The CI failed at XML-SAX-Expat, "Can't locate XML/Parser.pm in INC (you may need to install the XML::Parser module)"

PR #4478, opened 2024-10-18
nftables: New module, provides the "nft" utility

dagbrown:
Note that this release of nftables requires the very latest libnftnl (1.2.8) over in core which currently has a PR waiting for verification, so this should probably not be merged until that one is ready.


esselfe:
Retest this please.

esselfe:
Oh, this time (2nd) it failed because libnftnl is already installed, so it won't pick the most recent one... the CI will rebuild an image tonight then it should be fine :P

PR #4498, opened 2024-10-30
firefox: version bumped to 132.0

esselfe:


esselfe:
The CI failed because XML-SAX-Expat can't find the Parser perl module.

PR #4554, opened 2024-11-14
alsa-utils: version bumped to 1.2.13

esselfe:


esselfe:
The CI failed because alsa-lib 1.2.13 is not yet available.

PR #4557, opened 2024-11-16
neomutt: version bumped to 20241114

esselfe:


esselfe:
The CI failed at w3m, but this one succeeds here...

esselfe:
Retest this please.

esselfe:
It fails with this error:

x11_w3mimg.c:16:10: fatal error: glib-object.h: No such file or directory

PR #4607, opened 2024-11-29
OpenBLAS: A test.

stumbles:


stumbles:
Certainly a CI issue. Locally the compile log does not have any of those -W errors about parenthesis, fgets, etc CI is calling out.

stumbles:
There it is;  

` 1049 |                                         printf(" ******* FATAL ERROR - PARAMETER NUMBER %d WAS CHANGED INCORRECTLY *******\n",i__);`

Why locally that isn't an issue, but is with CI, I guess......... only its image knows.

PR #4613, opened 2024-12-01
CPU-X: version bumped to 5.1.1

esselfe:


esselfe:
This CI fails with this error:

CMake Error at data/CMakeLists.txt:41 (if):
  if given arguments:

    "STREQUAL" ""

  Unknown arguments specified


-- Configuring incomplete, errors occurred!

I don't get that here.

PR #4701, opened 2024-12-13
neomutt: version bumped to 20241212

esselfe:


esselfe:
The CI failed at w3m with:

x11_w3mimg.c:16:10: fatal error: glib-object.h: No such file or directory

PR #4765, opened 2024-12-24
openafs: Bump to 1.8.13.1.

stumbles:


stumbles:
Retest this please.

esselfe:
Here I tested it and it only complains because the secondary sources are not numbered linearly. With SOURCE{,2,3,5} it fails to download the patch, but with SOURCE{,2,3,4} it works.

PR #4806, opened 2024-12-29
XML-SAX: fix first-time installs

esselfe:
For some reason, during the installation phase, it fails with missing SAX.pm. Note that it succeeds if it's already installed, like during updates only.


esselfe:
The Lunar perl build function uses a destdir approach, so this might be why.

lunar-linux-b01:
SUCCESS!

PR #4810, opened 2024-12-29
astyle: Bump to 3.6.6.

stumbles:


stumbles:
CI doesn't pick `depends %` so astyle will tank missing a java. Merging anyway.

PR #4867, opened 2025-01-03
foot: version bumped to 1.20.1

esselfe:


lunar-linux-b01:
SUCCESS!

Ratler:
@esselfe I noticed one issue with this module. It installs systemd services, but they are not properly tracked by lunar. The module requires an additional file.

esselfe:
I have the user services logged in the foot install logs here...

PR #4902, opened 2025-01-07
ablc: Bump to 1.9.2.

stumbles:


stumbles:
Retest this please

stumbles:
AFAIK, CI will never pass this since ablc depends on apache-ant, which depends on %JAVA_SDK and CI doesn't know how to deal with the variable %MODULE in depends.

Correct me if wrong.

PR #4924, opened 2025-01-09
zlib-ng: Adding. Compatibility mode left disabled since it would

stumbles:
over write a couple of headers and zlib.pc from core. This way no zlib files are trampled.


lunar-linux-b01:
SUCCESS!

PR #4925, opened 2025-01-09
minizip-ng: Adding. Comaptibility is left disabled so it does not

stumbles:
over write a pc file from minizip.


lunar-linux-b01:
SUCCESS!

PR #4961, opened 2025-01-11
python-pycairo: This needs to use meson so py3cairo.h gets installed

stumbles:
which is needed by meld PR4954.


lunar-linux-b01:
SUCCESS!

PR #4987, opened 2025-01-13
tagparser: build shared library instead of static

esselfe:
Forgot to add PRE_BUILD in previous PR :P


lunar-linux-b01:
SUCCESS!

PR #5015, opened 2025-01-17
cdk: version bumped to 5.0-20250116

esselfe:


lunar-linux-b01:
SUCCESS!

esselfe:
With libtool it installs into /usr/lib/.libs, so disable libtool so it's installed in the right place. I haven't found something to change in configure.in nor Makefile.in.

PR #5153, opened 2025-02-05
hwdata: version bumped to 0.392

esselfe:
Move the .pc file from /usr/share/pkgconfig to /usr/lib/pkgconfig.


lunar-linux-b01:
SUCCESS!

PR #5173, opened 2025-02-09
rsyslog: fix build with newer libgcrypt

esselfe:
The error I had in configure was:

configure: error: libgcrypt-config not found in PATH

Which has been removed from recent libgcrypt.


lunar-linux-b01:
SUCCESS!

lunar-linux-b01:
SUCCESS!

PR #5224, opened 2025-02-14
go-bootstrap: switch to prebuilt bootstrap (1.22.12)

esselfe:
Ratler suggested we stop introducing additional bootstraps. With Go 1.24.0, a 4th one was needed. So use the minimum required prebuilt version instead.


lunar-linux-b01:
SUCCESS!

PR #5227, opened 2025-02-14
transmission: add a Gentoo patch for newer miniupnpc

esselfe:
The error I was getting was:

/usr/src/transmission-4.0.6/libtransmission/port-forwarding-upnp.cc:278:29: error: too few arguments to function 'int UPNP_GetValidIGD(UPNPDev*, UPNPUrls*, IGDdatas*, char*, int, char*, int)'


lunar-linux-b01:
SUCCESS!

PR #5451, opened 2025-06-27
bind: version bumped to 9.20.10

esselfe:


esselfe:
I think the CI failed because it has the unpatched version of file, which doesn't read zip files properly.

stumbles:
Its not limited to CI. I have the same issue with **some** source urls but not all. Not been able to figure out the real issue.

esselfe:
Closing with ba6dc9913c54b594acb01b17c219fb46c726932c from florin.

PR #5454, opened 2025-06-28
mbedtls: specify cmake policy version minimum

esselfe:


esselfe:
I tried mbedtls 3.6.3.1, but it breaks bctoolbox 5.3.101

lunar-linux-b01:
SUCCESS!

PR #5470, opened 2025-07-04
php: version bumped to 8.4.10

esselfe:


esselfe:
The CI failed at libavif because rustc is not recent enough. Currently it's rustc 1.85.1 and it wants 1.86.

esselfe:
Closing with 062809b8d765cb65cd0fe78c50cbd83658f69eb1 from @florin65

PR #5487, opened 2025-07-06
conky: version bumped to 1.22.2

esselfe:


lunar-linux-b01:
SUCCESS!

esselfe:
The new vim syntax file is from https://github.com/vim-scripts/syntaxconkyrc.vim/blob/master/syntax/conkyrc.vim

I think the one in the source (conkyrc.vim.j2) has to be processed first, since the error seems to be about lua code, not vim syntax code, but I'm not too used to lua.

PR #5535, opened 2025-08-25
gkrellm: add a patch to restore network monitoring with linux 6.16.1-3

esselfe:
So starting with linux 6.16.1, gkrellm doesn't show networking usage anymore. Some Gentoo user said it was because rewind() was now returning ESPIPE as error.

The fix consists of reopening and reclosing the metrics /proc files on every call instead of using rewind().


lunar-linux-b01:
SUCCESS!

PR #5597, opened 2025-09-14
Sqlite debloat: remove tcl mandatory dependency, forced Python rebuild

dagbrown:
sqlite seems to have picked up some bloat along the way.

First of all, tcl is an OPTIONAL dependency, and should have never been made mandatory.

Second, rebuilding Python with sqlite support is also OPTIONAL and that should also be a choice.


stumbles:
When sqlite is upgraded it changes soname versions breaking python requiring a lunar fix.

stumbles:
The lunar fix of python/python2 no longer applies, so I removed the POST_INSTALL from the recent sqlite bump.

PR #6555, opened 2026-04-17
cups: Version bumped to 2.4.17

lunar-linux-b01:
Upgrade cups from version 2.4.16 to 2.4.17

---
*Automated PR created by n8n + Claude AI*

### DIFF
--- a/printer/cups/DETAILS
+++ b/printer/cups/DETAILS
@@ -1,12 +1,12 @@
-          MODULE=cups
-         VERSION=2.4.16
-          SOURCE=$MODULE-$VERSION-source.tar.gz
-      SOURCE_URL=https://github.com/OpenPrinting/cups/releases/download/v${VERSION}/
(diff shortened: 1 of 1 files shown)


esselfe:
Fixes those CVEs:

CVE-2026-27447: The scheduler treated local user and group names as case-
insensitive.
CVE-2026-34978: The RSS notifier could write outside the scheduler's RSS
directory.
CVE-2026-34980: The scheduler did not filter control characters from option
values.
CVE-2026-34979: The scheduler did not always allocate enough memory for a
job's options string.
CVE-2026-34990: The scheduler incorrectly allowed local certificates over the
loopback interface.
CVE-2026-39314: Fixed the range check for job password strings.
CVE-2026-39316: Fixed a printer subscription bug in the scheduler.
CVE-2026-NNNNN: Fixed a SNMP string conversion bug in the backends.

PR #7043, opened 2026-06-30
openttd: fix build

esselfe:


esselfe:
The make install command was failing, and due to a missing '&&' chaining the tar commands were also failing.

PR #7072, opened 2026-07-13
git: Fix the NO_RUST flag

dagbrown:
Seems it's changed from WITH_RUST to NO_RUST

Also they're threatening that with git 3.x, rust will be a required
dependency.


stumbles:
I thought logic wasn't allow in DEPENDS.

PR #7124, opened 2026-07-22
firefox: Version bumped to 153.0

lunar-linux-b01:
Upgrade firefox from 152.0.6 to 153.0

- Updated VERSION to 153.0
- Updated SOURCE_VFY to bc510f74c8c4c692d31e559aeb5850849d13bc98214ac81e004f66f819a55522
- Updated UPDATED date to 20260722

---
*Automated PR created by n8n + Claude AI*

### DIFF
--- a/web/firefox/DETAILS
+++ b/web/firefox/DETAILS
@@ -1,11 +1,11 @@
     MODULE=firefox
-   VERSION=152.0.6
+   VERSION=153.0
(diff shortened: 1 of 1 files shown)


esselfe:
The CI fails because of a cbindgen being too old, but when trying on my PC with that fixed the build is missing libfreeblpriv3.so library. Note that libfreebl3.so comes from the nss module, but I haven't found how to have the "priv" one generated.

